PR #11, opened 2020-06-10
Standardize the code

ItsDrike:
## Summary

This PR standardizes the code style to follow basic `flake8` rules and makes sure the code is properly type-hinted everywhere.
It also implements `black` for keeping the code style consistent

## Details

- Install `black`(which affects **`Pipfile`** and **`Pipfile.lock`**)
    - `black` is also added to `.pre-commit-config.yaml` for consistency
    - This means `black` will now be used to evaluate every new commit with **Travis-CI**
- Add Type-Hinting everywhere as described in [**`CONTRIBUTING.md`**](https://github.com/The-Codin-Hole/HotWired-Bot/blob/master/CONTRIBUTING.md)
- Splits [**`tools.py`**](https://github.com/The-Codin-Hole/HotWired-Bot/blob/aaabd1a83bded9a7248ed56a27e3dfbfc08a0f7b/cogs/tools.py) into itself and **`converters.py`** *(more details in 6c0770fd5b0bbdd43203a28e68046b493843a231)*
- Ensure to use `""` by default instead of `''`
- Remove all *"dead-code"*
- Make the code much more readable.

PR #22, opened 2020-06-11
Update events and Some code-compiler edits

janaSunrise:


cogs/codesandbox.py
```diff
@@ -21,6 +21,7 @@ def _clean_code(self, code):
             return '\n'.join(code.split('\n')[1:-1])
         return code.strip('`\n')
 
+    # TODO : try to restrict to verified users only.
     @commands.is_owner()
     @commands.command(name='eval', hidden=True)
     async def _eval(self, ctx, *, code: str):
```
janaSunrise, line 27: Can this function and it's content be more refined or not

cogs/utils/paginator.py
```diff
@@ -0,0 +1,302 @@
...
+        Whether to show an entry count in the footer.
+    Attributes
+    -----------
+    embed: discord.Embed
+        The embed object that is being used to send pagination info.
+        Feel free to modify this externally. Only the description,
+        footer fields, and colour are internally modified.
+    permissions: discord.Permissions
+        Our permissions for the channel.
+    """
```
janaSunrise, line 42: Can we remove this comment, or make it smaller?

janaSunrise (commented):
I have reviewed everything an left a comment if i have a query, Please reviem then again, and send me the changes if needed

cogs/utils/paginator.py
ItsDrike: That should be a part of #11 don't change the comment here.

cogs/events.py
```diff
@@ -1,27 +1,89 @@
...
 
     @commands.Cog.listener()
     async def on_message(self, message):
         if message.author.id == self.client.user.id:
             return
 
-        if message.content.lower().startswith("help"):
-            await message.channel.send("Hey! Why don't you run the help command with `>>help`")
+        if message.author.bot:  # make sure the bot doesn't respond to other bots
+            return
```
ItsDrike, line 23:
```py
       if message.author.id == self.client.user.id:
            return
```
This is checking if the author is the bot itself and returning earlier if ture.
```py
        if message.author.bot:  # make sure the bot doesn't respond to other bots
            return
```
This is doing the same thing, but this one is checking for all bots, not only self.

That means there is no point to keep both conditions, the second one covers the first one and also checks for other bots, meaning that one is enough, the first condition can be removed.

cogs/events.py
```diff
@@ -1,27 +1,89 @@
...
 
         await self.client.process_commands(message)
 
+    @commands.Cog.listener()
+    async def on_guild_join(self, guild: discord.Guild):
+
+        hw = self.client.get_user(715545167649570977)
+        logchannel = self.client.get_channel(704197974577643550)
```
ItsDrike, line 31: These IDs shouldn't be hard coded, maybe they can also be in `constants.py`.
Same goes for line 77

cogs/events.py
```diff
@@ -1,27 +1,89 @@
...
+
+        await logchannel.send(
+            f"The bot has been Removed From **{guild.name}** ! It Sucks! :sob: :sneezing_face: "
+        )
+
+
     @commands.Cog.listener()
     async def on_error(self, event, *args, **kwargs):
-        error_message = f"```py\n{traceback.format_exc()}\n```"
+        error_message = f"```\n{traceback.format_exc()}\n```"
```
ItsDrike, line 86: Why is this changed?

cogs/utils/constants.py
```diff
@@ -14,10 +16,66 @@
 admin_invite_link = "https://discord.com/api/oauth2/authorize?client_id=715545167649570977&permissions=8" \
                     "&redirect_uri=https%3A%2F%2Fdiscord.gg%2FvP26dCy&scope=bot "
 
+SUPPORT_SERVER = "https://discord.gg/CgH6Sj6"
+
 COMMAND_PREFIX = ">>"
 
 line_img_url = "https://cdn.discordapp.com/attachments/581139962611892229/692712698487767080/animated_line.gif"
 
+# Bot replies
```
ItsDrike, line 25:
```suggestion
```

It's fairly obvious what is this, no need for the comment

ItsDrike (changes requested):
This branch is removing `CONTRIBUTING.md` which is unacceptable, this needs to be reverted.
It is also changing `Pipfile.lock` which it shouldn't, this should also be reverted.

janaSunrise (commented):
so, as you can see, i cleaned the code as much possible, now i need to clean my director on my local system, it's bcome messy, i will pull the branch, which you say, and start the next branch work on it.

ItsDrike (changes requested):
You need to restore `CONTRIBUTING.md` and revert `Pipfile.lock` as these files shouldn't be affected by this PR.

cogs/events.py
```diff
@@ -32,11 +33,66 @@ async def on_error(self, event, *args, **kwargs):
...
+        if not constants.DEV_MODE:
+            # await self.error_hook.send(embed=em)
+            await self.client.send_message(embed=em)
         else:
             traceback.print_exc()
 
+    @commands.Cog.listener()
+    async def on_guild_join(self, guild: discord.Guild):
+
+        hw = self.client.get_user(constants.hotwired_id)
```
ItsDrike, line 45: I believe there is a way to get bot id directly. You should take a look at [`discord.Bot.user`](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html?highlight=bot#discord.ext.commands.Bot.user) it seems like it's a reference to bot which means you could simply do `self.client.user.id` instead of saving it in constants.
This ID is also mentioned on line 63

ItsDrike (changes requested):

cogs/events.py
janaSunrise: i will check it asap

cogs/events.py
```diff
@@ -32,11 +33,66 @@ async def on_error(self, event, *args, **kwargs):
...
+        if not constants.DEV_MODE:
+            # await self.error_hook.send(embed=em)
+            await self.client.send_message(embed=em)
         else:
             traceback.print_exc()
 
+    @commands.Cog.listener()
+    async def on_guild_join(self, guild: discord.Guild):
+
+        hw = self.client.get_user(self.client.user.id)
```
ItsDrike, line 45: Since `self.client.user` will return `discord.User` instance, there is no need to get ID from that and get the user from it using `self.client.get_user`. We can simply do:
```suggestion
        hw = self.client.user
```

ItsDrike (changes requested):

cogs/events.py
```diff
@@ -32,11 +33,66 @@ async def on_error(self, event, *args, **kwargs):
             title=event
         )
 
-        if not self.dev_mode:
-            await self.error_hook.send(embed=em)
+        if not constants.DEV_MODE:
+            # await self.error_hook.send(embed=em)
```
ItsDrike, line 37:
```suggestion
```
There shouldn't be any dead code

cogs/events.py
```diff
@@ -32,11 +33,66 @@ async def on_error(self, event, *args, **kwargs):
...
+
+        embed = discord.Embed(
+            title="Greetings",
+            description=textwrap.dedent(f"""
+                Thanks for adding HotWired in this server,
+                **HotWired** is a multi purpose discord bot that has Moderation commands, Fun commands, 
+                Music commands and many more!. 
+                The bot is still in dev so you can expect more commands and features.To get a list of commands , 
+                please use **{PREFIX}help** 
+            """),
```
ItsDrike, line 56: I'm not sure `textwrap.dedent` is the right way to do it here, because this function doesn't just remove the new lines, wherever you used new line there will actually be a new line, it will not get merged into a single line.

ItsDrike (changes requested):

cogs/events.py
janaSunrise: Then, we gotta Organize it!

ItsDrike: > Then, we gotta Organize it!

I'm not sure what that means, you simply can't use dedent here

cogs/utils/constants.py
```diff
@@ -2,13 +2,17 @@
 
 from discord import Color
 
+DEV_MODE = True
+
+owner_id = 688275913535914014
```
ItsDrike, line 7: What is this?

cogs/utils/constants.py
```diff
@@ -21,10 +25,65 @@
     "&redirect_uri=https%3A%2F%2Fdiscord.gg%2FvP26dCy&scope=bot "
 )
 
+SUPPORT_SERVER = "https://discord.gg/CgH6Sj6"
+
 COMMAND_PREFIX = ">>"
 
 line_img_url = "https://cdn.discordapp.com/attachments/581139962611892229/692712698487767080/animated_line.gif"
 
+NEGATIVE_REPLIES = [
```
ItsDrike, line 34:
```suggestion
# Magic 8-Ball responses
NEGATIVE_REPLIES = [
```

`NEGATIVE_REPLIES` alone isn't very clear we should either use a comment, as suggested or rename the variables to something like `BALL8_NEGATIVE_REPLIES`

ItsDrike (changes requested):

cogs/utils/constants.py
janaSunrise: My ID? I Changed it to a list of us 3 soon

ghost (changes requested):

ghost (approved):
Thanks for fulfilling me wishes 😊

PR #25, opened 2020-06-11
Add multiban and clear command

janaSunrise:
The moderation cog needs to be really cleaned, it is messy, as the code was added and edited in a hurry.


cogs/moderation.py
```diff
@@ -1,10 +1,33 @@
...
+
+async def resolve_member(guild, member_id):
+    member = guild.get_member(member_id)
+    if member is None:
+        if guild.chunked:
+            raise MemberNotFound()
+        try:
+            member = await guild.fetch_member(member_id)
+        except discord.NotFound:
+            raise MemberNotFound() from None
```
ItsDrike, line 26:
```suggestion
            raise MemberNotFound()
```
What is `from None` it doesn't seem necessary

cogs/moderation.py
```diff
@@ -1,10 +1,33 @@
...
+    return member
+
+# Moderation Main Class
```
ItsDrike, line 29: Fairly obvious, I don't see a need for this comment
```suggestion
```

cogs/moderation.py
```diff
@@ -43,6 +66,65 @@ async def ban(self, ctx, member: discord.Member, *, reason="No Reason Stated."):
...
+    @commands.command()
+    @commands.guild_only()
+    @commands.bot_has_permissions(ban_members=True)
+    @commands.has_permissions(ban_members=True)
+    async def multiban(self, ctx, members: commands.Greedy[MemberID], *, reason: ActionReason = None):
+        """Bans multiple members from the server.
+        This only works through banning via ID.
+        In order for this to work, the bot must have Ban Member permissions.
+        To use this command you must have Ban Members permission.
+        """
```
ItsDrike, line 106:
```suggestion
        """
        Bans multiple members from the server.
        
        This only works through banning via ID.
        In order for this to work, the bot must have Ban Member permissions.
        To use this command you must have Ban Members permission.
        """
```
Multiline comments should have 1 line (or more when needed) separated as header and remaining lines should be used as more detailed description.

cogs/moderation.py
```diff
@@ -43,6 +66,65 @@ async def ban(self, ctx, member: discord.Member, *, reason="No Reason Stated."):
...
+
+        if reason is None:
+            reason = f'Action done by {ctx.author} (ID: {ctx.author.id})'
+
+        total_members = len(members)
+        if total_members == 0:
+            return await ctx.send('Missing members to ban.')
```
ItsDrike, line 113:
```suggestion
            return await ctx.send('No members to ban.')
```
This is a bit more grammatically correct.

cogs/moderation.py
```diff
@@ -1,10 +1,33 @@
...
+class MemberNotFound(Exception):
+    pass
+
+async def resolve_member(guild, member_id):
+    member = guild.get_member(member_id)
+    if member is None:
+        if guild.chunked:
+            raise MemberNotFound()
+        try:
+            member = await guild.fetch_member(member_id)
```
ItsDrike, line 24: Why are we fetching members only, we should be able to ban users which aren't necessarily members too (meaning they aren't on the server). [`fetch_user`](https://discordpy.readthedocs.io/en/latest/api.html?highlight=fetch#discord.Client.fetch_user) will probably be a better alternative.

Note that you also need to change all occurrences of `Member` and replace it with `User` in messages, docstrings and variable names.

cogs/moderation.py
```diff
@@ -43,6 +66,65 @@ async def ban(self, ctx, member: discord.Member, *, reason="No Reason Stated."):
...
+            if len(ret) > 512:
+                reason_max = 512 - len(ret) + len(argument)
+                raise commands.BadArgument(f'Reason is too long ({len(argument)}/{reason_max})')
+            return ret
+
+
+    class MemberID(commands.Converter):
+        async def convert(self, ctx, argument):
+            try:
+                m = await commands.MemberConverter().convert(ctx, argument)
```
ItsDrike, line 82: `m` is a very undescriptive variable name.
Same goes for lines: 86, 95, 198

cogs/utils/formats.py
```diff
@@ -0,0 +1,10 @@
+class plural:
+    def __init__(self, value):
+        self.value = value
+    def __format__(self, format_spec):
+        v = self.value
+        singular, sep, plural = format_spec.partition('|')
+        plural = plural or f'{singular}s'
+        if abs(v) != 1:
+            return f'{v} {plural}'
+        return f'{v} {singular}'
```
ItsDrike, line 10: There should be a description (docstring) to describe what this does.

cogs/moderation.py
```diff
@@ -1,10 +1,33 @@
 import asyncio
+from collections import Counter
+from cogs.utils.embedHandler import failure, info, success
 
 import discord
-from cogs.utils.embedHandler import failure, info, success
 from discord.ext import commands
 
-
+# Converters
```
ItsDrike, line 8:
```suggestion
```
This comment doesn't seem to be necessary.

cogs/moderation.py
janaSunrise: I don't know Much, pls guide me abt this.

janaSunrise: i replaced member with user

cogs/moderation.py
```diff
@@ -102,6 +186,45 @@ async def promote(self, ctx, member: discord.Member, role: discord.Role):
...
+            if msg.author == ctx.me:
+                await msg.delete()
+                count += 1
+        return {'Bot': count}
+
+    async def _complex_cleanup_strategy(self, ctx, search):
+        prefixes = tuple(self.client.get_guild_prefixes(ctx.guild))  # thanks startswith
+
+        def check(m):
+            return m.author == ctx.me or m.content.startswith(prefixes)
```
ItsDrike, line 201: You've changed the `m` variable in lines 82, 86, 95, but not here and on line 204, these need to be changed too

janaSunrise: ok on it.

cogs/moderation.py
```diff
@@ -1,29 +1,57 @@
...
+async def resolve_member(self, guild: discord.Guild, member_id: int) -> discord.User:
+    user = guild.get_member(member_id)
+    if user is None:
+        if guild.chunked:
+            raise MemberNotFound()
+        try:
+            user = await guild.fetch_member(member_id)
+        except discord.NotFound:
+            raise MemberNotFound()
+    return user
```
ItsDrike, line 41: * Why is this taking `self` as an argument, this function isn't a part of any class.
* We should also use `bot.fetch_user(member_id)` as shown in the [docs](https://discordpy.readthedocs.io/en/latest/ext/commands/api.html?highlight=fetch_user#discord.ext.commands.Bot.fetch_user) this is much better as it will find even the users which aren't connected to the guild, which can be helpful in case user which admins wanted to ban just disconnected. This also renders `guild` parameter irrelevant and we should accept `bot` instead.
* We should rename this function to `resolve_user` because of that

The final function may look something like this, but note that you also need to change all references to that function and this might not work so don't just paste it in, check it first.
```py
async def resolve_user(bot: Bot, member_id: int) -> discord.User:
    user = await bot.fetch_user(member_id)
    if not user:
        raise MemberNotFound
    return user
```

cogs/moderation.py
```diff
@@ -1,29 +1,57 @@
...
 
-    def __init__(self, bot: Bot) -> None:
-        self.bot = bot
+
+# TODO : Clean everything up
+class Moderation(Cog):
+    def __init__(self, client: discord.Bot) -> None:
+        self.client = client
```
ItsDrike, line 47: According to standards, we should use `bot` not `client`:
```suggestion
    def __init__(self, bot: Bot) -> None:
        self.bot = bot
```
Also, why were you type hinting this with `discord.Bot` when you can just do `Bot`, you have it imported

cogs/moderation.py
```diff
@@ -1,29 +1,57 @@
...
 
     @command()
     @bot_has_permissions(kick_members=True)
     @has_permissions(kick_members=True)
-    async def kick(self, ctx: Context, member: discord.Member, *, reason: str = "No reason.") -> None:
+    async def kick(self, ctx: Context, member: discord.Member, *, reason: str = "No specific reason") -> None:
```
ItsDrike, line 52: Why limit ourselves with only using `discord.Member` when you created a custom converter for `MemberID`. Use that but note that you'll need to check if the user which comes from it is Member or not if he isn't a member, this should give an error, as we can't kick, non-connected users

cogs/moderation.py
```diff
@@ -51,11 +79,72 @@ async def ban(self, ctx: Context, member: discord.Member, *, reason: str = "No R
...
+    class ActionReason(Converter):
+        async def convert(self, ctx: Context, argument: str) -> str:
+            ret = f"{ctx.author} (ID: {ctx.author.id}): {argument}"
+
+            if len(ret) > 512:
+                reason_max = 512 - len(ret) + len(argument)
+                raise BadArgument(f"Reason is too long ({len(argument)}/{reason_max})")
+            return ret
+
+    class MemberID(Converter):
```
ItsDrike, line 91: As you'll now be using the new `resolve_user` instead of `resolve_member` (as suggested) you might want to rename this to `UserID` instead

cogs/moderation.py
```diff
@@ -51,11 +79,72 @@ async def ban(self, ctx: Context, member: discord.Member, *, reason: str = "No R
...
+        async def convert(self, ctx: Context, argument: str) -> t.Union[str, type]:
+            try:
+                member = await MemberConverter().convert(ctx, argument)
+            except BadArgument:
+                try:
+                    member_id = int(argument, base=10)
+                    member = await resolve_member(ctx.guild, member_id)
```
ItsDrike, line 98:
```suggestion
        async def convert(self, ctx: Context, member_id: int) -> t.Union[str, type]:
            try:
                member = await MemberConverter.convert(ctx, argument)
            except BadArgument:
                try:
                    member = await resolve_member(ctx.guild, member_id)
```
We can simply accept `member_id` as `int` straight away, no need to convert it from string.
In case you'd really want to convert it from `str`, you also shouldn't specify `base=10` as that is the default value anyway.

cogs/moderation.py
```diff
@@ -51,11 +79,72 @@ async def ban(self, ctx: Context, member: discord.Member, *, reason: str = "No R
...
+                except ValueError:
+                    raise BadArgument(f"{argument} is not a valid member or member ID.") from None
+                except MemberNotFound:
+                    # hackban case
+                    return type("_Hackban", (), {"id": member_id, "__str__": lambda s: f"Member ID {s.id}"})()
+
+            if not can_execute_action(ctx, ctx.author, member):
+                raise BadArgument("You cannot do this action on this user due to role hierarchy.")
+            return member
```
ItsDrike, line 107: You might want to create a decorator for this instead of checking it in `MemberID` as that doesn't really make sense, this should be a converter not a check.

cogs/moderation.py
```diff
@@ -51,11 +79,72 @@ async def ban(self, ctx: Context, member: discord.Member, *, reason: str = "No R
...
+                failed += 1
+
+        await ctx.send(f"Banned {total_members - failed}/{total_members} members.")
+
+    @command()
+    @bot_has_permissions(ban_members=True)
+    @has_permissions(ban_members=True)
+    async def unban(self, ctx: Context, *, member: str) -> None:
         """Unban a User."""
+
```
ItsDrike, line 147: Why is there a blank line?
```suggestion
```

cogs/moderation.py
```diff
@@ -51,11 +79,72 @@ async def ban(self, ctx: Context, member: discord.Member, *, reason: str = "No R
...
+                except MemberNotFound:
+                    # hackban case
+                    return type("_Hackban", (), {"id": member_id, "__str__": lambda s: f"Member ID {s.id}"})()
+
+            if not can_execute_action(ctx, ctx.author, member):
+                raise BadArgument("You cannot do this action on this user due to role hierarchy.")
+            return member
+
     @command()
+    @guild_only()
```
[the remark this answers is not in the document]
janaSunrise: we might edit it later, leave it for now. I have a file `checks.py` in another PR, maybe we can edit it then.

cogs/moderation.py
janaSunrise: we can create a file for decorators later. pls, approve it for now.

cogs/moderation.py
janaSunrise: idk this

cogs/moderation.py
```diff
@@ -1,28 +1,53 @@
...
+        raise MemberNotFound()
+    return user
+
+
+# TODO : Clean everything up
 class Moderation(Cog):
     """This cog provides moderation commands."""
 
-    def __init__(self, bot: Bot) -> None:
+    def __init__(self, bot: discord.Bot) -> None:
```
ItsDrike, line 44: Why did you do this? what's wrong with `bot: Bot` there is no point in using `discord.Bot` it's the same but longer.

cogs/moderation.py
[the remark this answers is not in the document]
ItsDrike: No, there is nothing complicated to this, `checks.py` doesn't need to be involved here. This is just a separate function that can be added to your `Cog` class. It must be named `cog_check` and accept `ctx: Context`. This function will be used as a global check for the cog, I don't see a reason to wait for `checks.py`.

cogs/moderation.py
```diff
@@ -120,6 +196,45 @@ async def promote(self, ctx: Context, member: discord.Member, role: discord.Role
...
+        async for msg in ctx.history(limit=search, before=ctx.message):
+            if msg.author == ctx.me:
+                await msg.delete()
+                count += 1
+        return {"Bot": count}
+
+    async def _complex_cleanup_strategy(self, ctx: Context, search: int) -> Counter:
+        prefixes = tuple(self.bot.get_guild_prefixes(ctx.guild))  # thanks startswith
+
+        def check(member: discord.Bot) -> bool:
```
ItsDrike, line 210: * Why not just use `member: Bot` instead of using `discord.Bot` which is the same but longer.
* Also why the hell are you setting `member` equal to `Bot` type? name it `bot` so that it's obvious what this parameter is, by naming it `member` everyone would expect a  `discord.Member` not `Bot` ...
* And `check` is an undescriptive name for a function, it's not acceptable, what is it checking?

cogs/moderation.py
ItsDrike: I'm not sure what's hard on that, just change type from `discord.Member` to `UserID` and check if the resulting member in `member` is a part of that guild.

cogs/moderation.py
```diff
@@ -20,91 +129,216 @@ def __init__(self, bot: Bot) -> None:
...
+            embed = Embed(
+                title="Insufficient permissions",
+                description="You can give someone role which is higher in the role hierarchy than your top role.",
+                color=Color.red(),
+            )
             await ctx.send(embed=embed)
             return
         if role in member.roles:
-            embed = Embed(title="Error", description=f"{member.mention} already had role {role.mention}!", color=Color.red(),)
+            embed = Embed(title="Error", description=f"{member.mention} already have {role.mention} role!", color=Color.red())
```
Theodore-Robinson, line 309: Small thingy, 
```suggestion
            embed = Embed(title="Error", description=f"{member.mention} already has the {role.mention} role!", color=Color.red())
```
Small thing, change the `already have` to `already has`

cogs/moderation.py
```diff
@@ -20,91 +129,216 @@ def __init__(self, bot: Bot) -> None:
...
+            await ctx.send(f"Sorry {ctx.author.mention}", embed=embed)
+            return
+
+        server_embed = discord.Embed(
+            title="User Kicked",
+            description=textwrap.dedent(
+                f"""
+                **Reason**: {reason}
+                **User**: {member.mention} (`{member.id}`)
+                **Actor**: {ctx.author.mention} (`{ctx.author.id}`)
```
janaSunrise, line 157:
```suggestion
                **Moderator**: {ctx.author.mention} (`{ctx.author.id}`)
```

cogs/moderation.py
```diff
@@ -20,91 +129,216 @@ def __init__(self, bot: Bot) -> None:
...
-        embed.set_thumbnail(url=member.avatar_url)
-        embed.set_author(name=member.name, url=member.avatar_url)
-        embed.set_footer(text=member.guild.name, icon_url=member.guild.icon_url)
+        server_embed = discord.Embed(
+            title="User Banned",
+            description=textwrap.dedent(
+                f"""
+                **Reason**: {reason}
+                **User**: {member.mention} (`{member.id}`)
+                **Actor**: {ctx.author.mention} (`{ctx.author.id}`)
```
janaSunrise, line 194:
```suggestion
                **Moderator**: {ctx.author.mention} (`{ctx.author.id}`)
```

cogs/moderation.py
```diff
@@ -20,91 +129,216 @@ def __init__(self, bot: Bot) -> None:
...
 
-        await ctx.send(embed=embed)
-        await member.send(embed=embed)
+        dm_embed = discord.Embed(
+            title="You were Banned",
+            description=textwrap.dedent(
+                f"""
+                {reason}
+
+                *Server: {ctx.guild.name}*
```
janaSunrise, line 208:
```suggestion
                **Server: {ctx.guild.name}**
```

cogs/moderation.py
```diff
@@ -20,91 +129,216 @@ def __init__(self, bot: Bot) -> None:
...
+        )
+        server_embed.set_thumbnail(url=member.avatar_url_as(format="png", size=256))
+
+        dm_embed = discord.Embed(
+            title="You were Kicked",
+            description=textwrap.dedent(
+                f"""
+                {reason}
+
+                *Server: {ctx.guild.name}*
```
janaSunrise, line 171:
```suggestion
                **Server: {ctx.guild.name}**
```

cogs/moderation.py
```diff
@@ -20,91 +129,216 @@ def __init__(self, bot: Bot) -> None:
...
+        """Unban a User."""
+        try:
+            await ctx.guild.unban(user)
+
+            embed = discord.Embed(
+                title="User Unbanned",
+                description=textwrap.dedent(
+                    f"""
+                    **User**: {user.mention} (`{user.id}`)
+                    **Actor**: {ctx.author.mention} (`{ctx.author.id}`)
```
janaSunrise, line 256:
```suggestion
                    **Moderator**: {ctx.author.mention} (`{ctx.author.id}`)
```

cogs/moderation.py
```diff
@@ -20,91 +129,216 @@ def __init__(self, bot: Bot) -> None:
...
+        except NotFound:
+            embed = discord.Embed(
+                title="No such ban",
+                description=textwrap.dedent(
+                    f"""
+                    There are no active bans on discord for {user.mention}.
+                    He isn't banned here.
+                    """
+                ),
+                color=discord.Color.red(),
```
janaSunrise, line 273:
```suggestion
                title="Ban not Found!",
                description=textwrap.dedent(
                    f"""
                    There are no active bans on discord for {user.mention}. He isn't Banned.
                    """
                ),
                color=discord.Color.dark_gold(),
```

cogs/moderation.py
ItsDrike: No, I like this to only be italic, not bold, that's bringing too much attention on Server and doesn't look good, I've tried it. The reason should be the important factor in the embed, not the server.

cogs/moderation.py
ItsDrike: > No, I like this to only be italic, not bold, that's bringing too much attention on Server and doesn't look good, I've tried it. The reason should be the important factor in the embed, not the server.

(Same as above)

cogs/moderation.py
ItsDrike: Fixed in cae3d0229dfe10db6f1ebd9ecdde6bc98c5f61d7

cogs/moderation.py
ItsDrike: Fixed in cae3d0229dfe10db6f1ebd9ecdde6bc98c5f61d7

cogs/moderation.py
ItsDrike: Fixed in cae3d0229dfe10db6f1ebd9ecdde6bc98c5f61d7

cogs/moderation.py
ItsDrike: Fixed in f48e1f8e5bb27c34e5949709170270bbb0fe5fd0

PR #28, opened 2020-06-12
Add hacking commands

janaSunrise:
added ip lookup. Next we have to add whois and fsociety tools too for pentesting, but we need a disclaimer too since, it can be dangerous, ok?
FSOCIETY TOOLS ARE FOUND HERE => https://github.com/Manisso/fsociety


cogs/hacking.py
```diff
@@ -0,0 +1,76 @@
...
+                    color=discord.Color.dark_green()
+                )
+                embed.add_field(name="**❯❯ Locational Data**", value=loc_data, inline=False)
+                embed.add_field(name="**❯❯ Basic Data**", value=basic_data, inline=False)
+                embed.add_field(name="**❯❯ Online Data**", value=online_data, inline=False)
+
+                await ctx.send(embed=embed)
+
+
+            except:
```
ItsDrike, line 57: Bare except is unacceptable

cogs/hacking.py
```diff
@@ -0,0 +1,76 @@
...
+                region = data['region']
+                country = data['country']
+                country_capital = data['country_capital']
+                country_tld = data['country_tld']
+                postal = data['postal']
+                utc_offset = data['utc_offset']
+                currency = data['currency']
+                languages = data['languages'].split(',')
+                asn = data['asn']
+                org = data['org']
```
ItsDrike, line 26: Too many variables here, I'd just use `data['city']` in the f-strings directly, no need to define so many varaibles which are only used once anyway

janaSunrise:
> As suggested before on some pull requests, this code also needs type-hints with or without the typing library and it also requires docstrings below class definitions and function definitions. Just a small request, everything else is fine.

The Types hints are being added by @ItsDrike , i will add the docstrings

ItsDrike:
> The Types hints are being added by @ItsDrike , i will add the docstrings

Even though I will be adding type hinting everywhere in #11 it would help if you implemented it on most of the new things, it will reduce the time I'll need for standardizing the code @janaSunrise

janaSunrise:
> > The Types hints are being added by @ItsDrike , i will add the docstrings
> 
> Even though I will be adding type hinting everywhere in #11 it would help if you implemented it on most of the new things, it will reduce the time I'll need for standardizing the code @janaSunrise

IDK how to implement type hints

ItsDrike:
This PR will be closed due to inactivity from authors side

PR #34, opened 2020-06-13
add hangman

janaSunrise:


cogs/games.py
```diff
@@ -58,6 +60,144 @@ def is_lucky():
...
+                   |      
+                   |     
+                   -
+                """
+            ]
+            return stages[tries]
+
+        def get_word():
+            word = random.choice(word_list)
+            return word.upper()
```
ItsDrike, line 141: There is no need for this to be a function, it's only referenced once in the code at line: 198, where you can simply use `word=random.choice(word_list).upper()`

cogs/games.py
```diff
@@ -58,6 +60,144 @@ def is_lucky():
...
+                await ctx.send(display_hangman(tries))
+                await ctx.send(word_completion)
+                await ctx.send("\n")
+            if guessed:
+                await ctx.send("Congrats, you guessed the word! You win!")
+            else:
+                await ctx.send("Sorry, you ran out of tries. The word was " + word + ". Maybe next time!")
+
+        word = get_word()
+        play(word)
```
ItsDrike, line 200: No need to make `play` function inside another function, you can just remove that and use it directly

janaSunrise: pls provide the suggestion

cogs/games.py
janaSunrise: ok

cogs/games.py
ItsDrike: what suggestion just put `word = random.choice(word_list).upper()` on top and remove `def play` so that it's not in function, suggestion would be huge in here and there is no need for it

janaSunrise: done actually.

cogs/games.py
```diff
@@ -40,15 +41,16 @@ async def roll(self, ctx: Context, min_limit: int = 1, max_limit: int = 10) -> N
...
             await ctx.send(embed=embed)
         else:
-            embed = Embed(title="Random Roll", color=Color.red(), description="Please specify numbers with difference of **at least 2**",)
+            embed = Embed(title="Random Roll", color=Color.red(), description="Please specify numbers with difference of **at least 2**")
             await ctx.send(embed=embed)
 
     @command(aliases=["8ball"])
     async def ball8(self, ctx: Context, *, question: str) -> None:
         """Play 8ball."""
+        answer = ""
```
ItsDrike, line 53: Is this really needed? `answer` should always get set by one of the conditions above

cogs/games.py
```diff
@@ -58,10 +60,151 @@ async def ball8(self, ctx: Context, *, question: str) -> None:
...
-        embed = Embed(title="Magic 8-ball", color=Color.blurple(),)
+        embed = Embed(title="Magic 8-ball", color=Color.blurple())
         embed.add_field(name="Question", value=question)
         embed.add_field(name="Answer", value=answer)
 
+    @command()
+    async def hangman(self, ctx: Context) -> None:
+        def display_hangman(tries: int) -> str:
+            stages = [  # final state: head, torso, both arms, and both legs
+                r"""
```
ItsDrike, line 71: This comment should go above r""" not besides the bracket
```suggestion
            stages = [  
                # final state: head, torso, both arms, and both legs
                r"""
```

cogs/games.py
```diff
@@ -58,10 +60,151 @@ async def ball8(self, ctx: Context, *, question: str) -> None:
...
+        while not guessed and tries > 0:
+            await ctx.send("Please guess a letter or word: ")
+            input = await self.bot.wait_for("message", check=check)
+            guess = input.upper()
+
+            if len(guess) == 1 and guess.isalpha():
+                if guess in guessed_letters:
+                    await ctx.send(embed=Embed(description=f"You already guessed the letter {guess}", color=Color.blurple()))
+                elif guess not in word:
+                    await ctx.send()
```
ItsDrike, line 167: Why send an empty message here?
Same goes for Line 203

cogs/games.py
```diff
@@ -58,10 +60,151 @@ async def ball8(self, ctx: Context, *, question: str) -> None:
...
+        word_completion = "_" * len(word)
+        guessed = False
+        guessed_letters = []
+        guessed_words = []
+        tries = 6
+
+        await ctx.send(embed=Embed(title="Let's play Hangman!", color=Color.dark_magenta()))
+        await ctx.send(embed=Embed(title="Hangman Status", description=display_hangman(tries), color=Color.dark_magenta()))
+        await ctx.send(embed=Embed(title="Word Completion", description=word_completion, color=Color.dark_magenta()))
+        await ctx.send("\n")
```
ItsDrike, line 156: Why are you sending an empty line here?
Same goes for Line 198

PR #36, opened 2020-06-13
Add checks

janaSunrise:
## Summary

This PR aims to add multiple checks in the form of functions.
These checks should make checking certain things easier by only using one of these functions instead of writing a custom check every time.

## TODO Checks

* [x] Check owner or not
* [x] Check if a user has greater perms or not
* [x] Check if a user has higher roles of not
* [x] Check if a user can Kick
* [x] Check if a user can Ban
* [x] Check if a user can manage messages
* [x] Check if a User can Manage Channels
* [x] Check if a User can Manage Server


assets/checks.py
```diff
@@ -0,0 +1,28 @@
+from cogs.utils.constants import ids
+
+
+async def perms(perms, g, c, m):
```
ItsDrike, line 4: What are `g`, `c` and `m`, you couldn't have chosen more undescriptive names.
This is unacceptable, these must be changed.

Same goes for lines: 15 and 21

cogs/utils/constants.py
```diff
@@ -2,6 +2,11 @@
 
 from discord import Color
 
+ids = [
```
ItsDrike, line 5: `ids` of what, undescriptive name, change this to something more specific

ItsDrike (changes requested):
It's a bit weird that you've added **`status: needs review`** on **Draft** PR, but ok, here is the review.

You're not using any **Type Hinting** which is mandatory to keep our code clean and consistent as described in [**`CONTRIBUTING.md`**](https://github.com/The-Codin-Hole/HotWired-Bot/blob/aaabd1a83bded9a7248ed56a27e3dfbfc08a0f7b/CONTRIBUTING.md). 

Other than that and few minor suggestions I guess it's fine, although it's very uncommon to request a review on Draft PRs directly with **`status: needs review`**.

PR #39, opened 2020-06-15
Change mathjs and wolfram

janaSunrise:
* [x] add mathjs
* [x] add wolfram
* [x] add ncatlab


ghost (changes requested):
Awaiting till errors are solved and raw image sending to embed image sending.
Will do it together (me and jana)

cogs/study.py
```diff
@@ -14,43 +26,295 @@ async def pcalc(self, ctx):
...
+
+        temp_msg = await ctx.reply("Sending query, please wait.")
+
+        appid = s.WOLFRAM_APPID
+
+        try:
+            result = await self.get_query(query, appid)
+        except Exception:
+            await ctx.send("An unknown exception occurred while fetching query.")
+            return
```
ItsDrike, line 236: This is generally not a good practice, can **any** exception occur here? 
If not you should only except those you want to except so in case some exception which we didn't realize could happen occurs here, we wouldn't know about it as it would get handled by this.

cogs/study.py
```diff
@@ -14,43 +26,295 @@ async def pcalc(self, ctx):
...
+        asyncio.ensure_future(ctx.offer_delete(out_msg))
 
-        embed = discord.Embed(title="Question Results")
+        if extra:
+            try:
+                await ctx.bot.add_reaction(out_msg, ctx.bot.objects["emoji_more"])
+            except discord.Forbidden:
+                pass
+            except discord.HTTPException:
+                pass
```
ItsDrike, line 283: Use `contextlib.suppress` instead:
```py
from contextlib import suppress


with suppress(discord.Forbidden, discord.HTTPException):
    await ctx.bot.add_reaction(out_msg, ctx.bot.objects["emoji_more"])
```
This is generally a preferred way to do it as `try-except` block should only be used when you want to do something in case exception occurs, when you're only using it with `pass`, `suppress` is a better alternative.

ItsDrike (changes requested):
You also need to add type hinting to the end of your functions, to show us what is their return type

PR #41, opened 2020-06-16
Add dotenv support

janaSunrise:


cogs/sudo.py
```diff
@@ -43,6 +44,20 @@ async def sudo(self, ctx: Context) -> None:
             embed = Embed(description="Invalid sudo Command Passed!", color=Color.red())
             await ctx.send(embed=embed)
 
+    async def check_if_is_owner(self, ctx: Context) -> t.Union[bool, None]:
+        if ctx.author.id in [710400991761137666, 688275913535914014, 306876636526280705]:
```
ItsDrike, line 48: These IDs should be part of `constants.py` same goes for line 56

cogs/sudo.py
```diff
@@ -6,7 +6,9 @@
 import psutil
 from discord import Color, Embed
 from discord import __version__ as discord_version
-from discord.ext.commands import Bot, Cog, Context, group
+from discord.ext.commands import Bot, Cog, Context, check, group
+import typing as t
+from .utils import constants
```
ItsDrike, line 11: Please sort all imports, vscode has a built-in way to do this:
Peek 2020-06-16 10-55

PR #43, opened 2020-06-18
Added a working paginator

ghost:


utils/paginator.py
```diff
@@ -0,0 +1,236 @@
...
+        Timeout for the paginator to exit.
+        If reached, it tells the paginator to exit and clear all the reactions.
+        Defaults to 300.0s (5 minutes)
+    kwargs: Union[str, discord.Emoji]
+        Key-Word arguments to override default reaction emojis
+        Can be used like so:
+
+        `
+        await Paginator().send(timeout=400.0, first_page_emoji='👆', next_page_emoji='✌️')
+        `
```
janaSunrise, line 123:
```suggestion
  
```

utils/paginator.py
```diff
@@ -0,0 +1,236 @@
...
+    paginator: commands.Paginator
+        A static paginator which is pre-configured with the entries passed.
+        It is not recommended to manually process this paginator unless if
+        necessary, use instance.send() instead.
+    embed: discord.Embed
+        Use this to change embed attributes such as color, title etc
+        Do note that things like footer (depending on the show_entry_count parameter) and description
+        will be overridden if specified.
+        Use it like so:
+
```
janaSunrise, line 49:
```suggestion
    

```

utils/paginator.py
```diff
@@ -0,0 +1,236 @@
...
+            Pages = Paginator()
+            Pages.embed = discord.Embed(
+                title='Example',
+                color='discord.Color.blurple()'
+            )
+        `
+    Raises
+    -----------
+    CannotPaginate: Exception
+        Raised when the bot lacks a permission/condition to display/process the paginator
```
janaSunrise, line 60:
```suggestion
     
```

utils/paginator.py
```diff
@@ -0,0 +1,236 @@
...
+        ]
+
+        for permission in mandatory_permissions:
+            if not permission:
+                raise CannotPaginate(
+                    f"""
+            Bot lacks a mandatory permission to paginate:\n\t
+            Permission: {permission}\n\t
+            Server: {self.guild.name} ({self.guild.id})\n\t
+            Command & Author: {ctx.command.name} ({ctx.author.id}) & {ctx.author.name}
```
janaSunrise, line 94:
```suggestion
            Command **{ctx.command.name}** invoked by {ctx.author.name} ({ctx.author.id})
```

utils/paginator.py
```diff
@@ -0,0 +1,236 @@
...
+        if self.show_page_length:
+            embed.set_footer(text=f"Page {page + 1}/{len(paginator.pages)}")
+
+        # Reaction emojis. Can be overridden by kwargs.
+        reaction_emojis = {
+            "first_page_emoji": "⏮️",  # First page
+            "prev_page_emoji": "⬅️",  # Previous Page
+            "next_page_emoji": "➡️",  # Next Page
+            "last_page_emoji": "⏭️",  # Last Page
+            "exit_emoji": "<:trashcan:722035312238526505>",  # Stop paginator and delete message (trashcan)
```
janaSunrise, line 143: Add an Emoji to Get the info about the pagination, and an Emoji to go to a valid page chose by me.

janaSunrise (changes requested):
Change as requested to get approved.

utils/paginator.py
Theodore-Robinson: Why delete this?

utils/paginator.py
Theodore-Robinson: Okay

utils/paginator.py
```diff
@@ -0,0 +1,191 @@
...
+from discord.ext import commands
+import typing as t
+from discord.ext.commands import Bot, Context
+
+
+class CannotPaginate(Exception):
+    """
+    Base Exception Class for Custom error 'Cannot Paginate'
+    """
+
```
janaSunrise, line 13:
```suggestion

```

utils/paginator.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    Implements a paginator that queries the user for the
+    pagination interface.
+    Pages are 1-index based, not 0-index based.
+    If the user does not reply within 2 minutes then the pagination
+    interface exits automatically.
+
+    
+
+
```
janaSunrise, line 27:
```suggestion

```

utils/paginator.py
```diff
@@ -0,0 +1,191 @@
...
+        self.paginator = paginator = BasePaginator(max_size=self.max_size, prefix="", suffix="")
+
+        for entry in entries:
+            paginator.add_line(entry)
+
+    async def send(self, timeout: t.Union[int, float] = 300.0, **kwargs: t.Union[str, discord.Emoji],) -> None:
+        """
+    Sends a paginator and processes it according to the reactions passed by the author.
+
+
```
janaSunrise, line 78:
```suggestion

```

janaSunrise (changes requested):

utils/paginator.py
```diff
@@ -0,0 +1,191 @@
+import asyncio
+import discord
+from discord.ext.commands import Paginator as BasePaginator
+from discord.ext import commands
+import typing as t
+from discord.ext.commands import Bot, Context
```
ItsDrike, line 6:
```suggestion
import asyncio
import typing as t

import discord
from discord.ext import commands
from discord.ext.commands import Bot, Context
from discord.ext.commands import Paginator as BasePaginator
```
These imports need to be properly sorted
<details>
<summary>you can use vscode's functionality for it as seen in this GIF</summary>
<p>

Peek 2020-06-18 10-01
</p>
</details>

ItsDrike (dismissed):

utils/paginator.py
```diff
@@ -0,0 +1,191 @@
...
+    Implements a paginator that queries the user for the
+    pagination interface.
+    Pages are 1-index based, not 0-index based.
+    If the user does not reply within 2 minutes then the pagination
+    interface exits automatically.
+
+    
+
+     
+    """
```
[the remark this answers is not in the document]
Theodore-Robinson: Yeah, it was a conflict when merging [jana's suggestion](#discussion_r441967307)

ItsDrike:
Waiting for author of this PR ( @e-xe ).

e-xe:
> Waiting for author of this PR ( @e-xe ) to address remaining requested changes and merge master into this branch in order to get it up to date with the current version of the repository.

It is being worked on, and there will be a completely new paginator coming up.

ItsDrike:
If this is only a WIP PR, please convert it to a draft PR to avoid confusion, Open pull requests are the ones waiting for review or waiting for the author to address new requested changes, not for adding new things which are still unfinished

bot/__main__.py
```diff
@@ -37,9 +39,29 @@
...
+        colorama.init(autoreset=True)
+
+        raise TokenNotFoundError(
+            colorama.Fore.RED + "\n"
+            "Token is not found, are you sure\n"
+            "you are running this file through\n"
+            "pipenv and there is a .env file\n"
+            "containing a BOT_TOKEN key on your\n"
+            "current working directory?"
+        )
```
therealdaneel, line 67:
```suggestion
        raise TokenNotFoundError(
            colorama.Fore.RED + "\n"
            "Token not found.\n"
            "Are you running this file through pipenv and\n"
            "is there a .env file containing a BOT_TOKEN key\n"
            "in your current working directory?\n"
        )
```
minor improvement to the error message

bot/cogs/coding/reference.py
```diff
@@ -87,7 +92,7 @@ async def _git_main_ref(part: str, ctx: Context, text: str) -> Message:
     base_url = f"https://git-scm.com/docs/{part}{text}"
     url = urllib.parse.quote_plus(base_url, safe=";/?:@&=$,><-[]")
 
-    async with aiohttp.ClientSession() as client_session:
+    async with aiohttp.ClientSession as client_session:
```
janaSunrise, line 95: why did u remove the parentheses

bot/cogs/coding/reference.py
```diff
@@ -108,14 +113,15 @@ async def _git_main_ref(part: str, ctx: Context, text: str) -> Message:
...
+            await client_session.close()
             return await ctx.send(embed=embed)
 
 
 git_ref = partial(_git_main_ref, "git-")
 git_tutorial_ref = partial(_git_main_ref, "")
 
 
-async def sql_ref(ctx: Context, text: str) -> Message:
+async def sql_ref(bot: Bot, ctx: Context, text: str) -> Message:
```
janaSunrise, line 124: why r u adding bot as a Param everywhere?

janaSunrise (changes requested):

bot/cogs/coding/reference.py
e-xe: Its pretty easy to find out why by looking at the code.
But the reason is because the aiohttp session is inside of the bot.
The reason why we do this to ensure that the session is properly closed in the shutoff command.

bot/cogs/coding/reference.py
e-xe: sorry, will fix 😅

bot/cogs/sudo.py
```diff
@@ -44,8 +42,9 @@ def get_uptime(self) -> str:
         else:
             return f"{hours} hr, {minutes} mins, and {seconds} secs"
 
+    @staticmethod
     async def is_owner(ctx: Context) -> t.Union[bool, None]:
```
ItsDrike, line 46: Rename this to a more meaningful `is_dev` and use pure `bool` return, simply return `False` on the failed check
```suggestion
    async def is_dev(ctx: Context) -> bool:
```

bot/cogs/sudo.py
```diff
@@ -44,8 +42,9 @@ def get_uptime(self) -> str:
...
-        if ctx.author.id in config.devs:
+        if ctx.author.id in devs:
             return True
         else:
```
ItsDrike, line 49: This else is unnecessary, either the first if statement will run, ending the function, or if not, you can simply do what's in else, the `else` will always get triggered only when the first if doesn't pass so we might as well not use it.

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+import typing as t
+from textwrap import dedent
+
+import discord
+from discord import Embed
+from discord.abc import Messageable
+from discord.ext import commands
+from discord.ext.commands import Context, Paginator
+
+logger = logging.getLogger(__name__)
```
ItsDrike, line 12: We shouldn't use logging before #83 I know it's my PR and I'm working on it, but we shouldn't touch logging until that, either convert these logs to print statements or wait for #83 to get merged.

Note that even after merging it, this would still be invalid as we'll be using `loguru` for logging, not the default `logging` as is done here

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+    """
+    Checks if the bot has required permissions to send or process
+    the paginator
+
+    Args:
+        channel (Messageable): The Context
+        bot (commands.Bot): The bot object, used for using bot.user if there is no guild
+
+    Returns:
+        bool: Whether the bot has all of the required permissions
```
ItsDrike, line 33: We don't use this doctring style, more in [`CONTRIBUTING.md/AutoDoc Formatting Directive`](https://github.com/therealdaneel/HotWired-Bot/blob/master/CONTRIBUTING.md#autodoc-formatting-directive)

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+            Brief Explanation: The arguments can be a list,
+            representing a page or it can be an embed.
+
+            Example:
+            Key-Value is required. Inline is optional since its
+            set to True as a default value.
+            To insert an embed (creating a new page),
+            create a new value on the list as shown below.
+
+            FieldPaginator([{'Page': '1', 'inline': True}, discord.Embed(title="Page 2")])
```
ItsDrike, line 79: This should be a docstring to `__init__` function, not the class itself, also `----------` serves as sufficient separation, there is no need to also indent the actual parameters mentioned

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+                _embed = discord.Embed()
+                for element in arg:
+                    if isinstance(element, dict):
+                        # If element is an empty dictionary
+                        if element == {}:
+                            raise PaginatorError(
+                                "FieldPaginator argument is a nested dictionary but does not have the mandatory "
+                                "key-value pair".strip()
+                            )
+                        elif len(element.items()) > 2:
```
ItsDrike, line 98: There is no need for `elif` here, we can use `if` since the purpose of `elif` is to only run if the first `if` won't pass, in this case, we don't need to check for that since if it will pass the function will raise an exception and therefore avoiding the need to check if the first if was run.

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+            Whether to make a new paginator page for every
+            entry.
+        prefix : str
+            A string that is added to the beginning to every
+            paginator page.
+        suffix : str
+            A string that is added to the end of every paginator
+            page.
+        max_size : int
+            Maximum amount of characters allowed per entry.
```
ItsDrike, line 148: This should be a docstring for `__init__`, not for the class itself

**This also applies to all occurrences of this issue**

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+    def __init__(
+        self,
+        entries: t.Optional[t.List[str]] = None,
+        individual_paging=True,
+        prefix='',
+        suffix='',
+        max_size=2000
+    ) -> None:
+
+        entries = entries if entries is not None else []
```
ItsDrike, line 159: While there is technically nothing wrong with your syntax, there is no need for ternary conditional here since it only resets the variable to itself or to the preferred value, using regular if statement would be preferred as it is more commonly understood hence making it easier to read and doesn't waste time on re-setting the variable to itself.
```suggestion
        if entries is None:
            entries = []
```

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+import discord
+from discord import Embed
+from discord.abc import Messageable
+from discord.ext import commands
+from discord.ext.commands import Context, Paginator
+
+logger = logging.getLogger(__name__)
+
+
+class PaginatorError(Exception):
```
ItsDrike, line 15: Naming this `PaginationError` makes more sense since it signifies some problem which occurs while paginating not an error of the paginator itself

Note: You'll also need to rename all references to this class
```suggestion
class PaginationError(Exception):
```

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+    embed = paginator.embeds[page]
+
+    # Reaction emojis. Can be overridden by kwargs.
+    reaction_emojis = {
+        "first_page_emoji": "⏮️",  # First page
+        "prev_page_emoji": "⬅️",  # Previous Page
+        "next_page_emoji": "➡️",  # Next Page
+        "last_page_emoji": "⏭️",  # Last Page
+        "exit_emoji": "<:trashcan:722035312238526505>",  # Stop paginator and delete message (trashcan)
+    }
```
ItsDrike, line 321: I'd like these to be on the top of this file to make them easily accessible and changeable, there should be nothing wrong with making this a global variable.

bot/core/paginator.py
```diff
@@ -0,0 +1,421 @@
...
+
+    for key in reaction_emojis.keys():
+        if key in [_ for _ in kwargs]:
+            reaction_emojis[key] = kwargs[key]
+
+    def update_footer(embed_: discord.Embed, page_: int) -> None:
```
ItsDrike, line 327: What's wrong with using `embed` and `page` why `embed_` and `page_`?

**Same applies for all functions with seemingly irrelevant underscores next to their arguments**
```suggestion
    def update_footer(embed: discord.Embed, page: int) -> None:
```

ItsDrike (changes requested):
This is just the initial review of the paginator where I considered individual problems, 

I'll need to look into the file itself as a whole and check if there can be some improvements made in the overall structure of this file, I will do this in later time.

bot/cogs/coding/reference.py
ItsDrike: I'm marking this resolved as I've referenced this same line in my requested change where I suggesting replacing this whole thing with `bot.aio_session` which seems more reasonable if, for some reason that will not be done, this should be re-opened

bot/core/bot.py
```diff
@@ -44,7 +51,13 @@ async def on_ready(self) -> None:
             )
             await self.log_channel.send(embed=embed)
 
-        print("Bot is ready")
+        print(colorama.Fore.GREEN + "Bot is ready")
+
+    async def shutoff(self) -> None:
```
ItsDrike, line 56: We should probably stick to `close` since that's the name in `Base_Bot` class and it is what things will call if they need to close the bot, that way if they call `close` it would fall back to the original `Base_Bot` instead of going here, since this doesn't override that function anymore, there is no real reason to rename this.

Note that by reverting this, you'll also need to revert renamed references to this function

ItsDrike (changes requested):
Don't forget to also address older requested changes, since this won't be adding paginator anymore, I marked all changes regarding paginator as resolved, but there are other changes which aren't connected to paginator, these need to get addressed together with this new one

PR #44, opened 2020-06-18
Renamed custom.py to common.py

DanDevDan:
Renamed the custom.py to common.py.


ItsDrike (changes requested):
This only removes **`cogs/custom.py`** it doesn't actually rename it.

SignalDevX:
It removes cogs/custom.py and replaces with cogs/common.py, with classes changed as well.

ItsDrike:
> It removes cogs/custom.py and replaces with cogs/common.py, with classes changed as well.

You probably forgot to stage it when committing because in **files changed** tab there is only the removal of **`cogs/custom.py`** and adjustment to **`main.py`**, but there are no changes or renaming to it.

image

SignalDevX:
You're right. Must of missed it.

PR #45, opened 2020-06-18
Renamed custom to common

DanDevDan:
Renamed custom to common and made the neccessary changes in main.py


ItsDrike (approved):
Umm, you could've just pushed another commit fixing it to #44, no need for new pull request, but ok.

It looks good, but I'll wait for @janaSunrise to also review this before we merge

janaSunrise (approved):

PR #47, opened 2020-06-19
Added docker and removed psutil & sudo.py since it was altering docker

ghost:
- [x] Added docker & .dockerignore
- [x] Removed the psutil module and sudo.py since they are not needed in my view and are altering docker


Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN pip install pipenv
+COPY Pipfile* ./
+RUN pipenv sync
+
+# Copying all the files files over to the workdir
+COPY . ./
+
+# Running the container when asked with pipenv
+ENTRYPOINT ["pipenv"]
+CMD ["run", "start"]
```
janaSunrise, line 18:
```suggestion
CMD ["run", "start"]

```

janaSunrise (changes requested):
I will close if you delete `sudo.py`, bc we have sudo commands like restart, shutdown, etc

ItsDrike:
> I will close if you delete `sudo.py`, bc we have sudo commands like restart, shutdown, etc

I actually agree here, we should keep `sudo.py` for these commands to control cogs and we might want to also keep `!sudo status` with some adjustments so it won't need `psutil` but the command is still useful to show how many servers is the bot running on, etc.

PR #48, opened 2020-06-19
Changes to main.py

Faholan:
Subclass bot, and better extension loading


main.py
```diff
@@ -15,102 +14,66 @@
...
-    bot.load_extension("cogs.tools")
-
-    bot.run(TOKEN)
-
-
-client.loop.create_task(change_status())
+extensions = ["cogs.codesandbox", "cogs.commands", "cogs.converters", "cogs.custom", "cogs.emotes",
+"cogs.events", "cogs.fun" ,"cogs.games", "cogs.infog", "cogs.moderation", "cogs.study", "cogs.sudo",
+"cogs.support", "cogs.tools"]
+#"cogs.coding"
```
janaSunrise, line 20:
```suggestion
extensions = [
"cogs.codesandbox", 
"cogs.commands", 
"cogs.converters", 
"cogs.custom", 
"cogs.emotes",
"cogs.events", 
"cogs.fun",
"cogs.games", 
"cogs.infog", 
"cogs.moderation", 
"cogs.study", 
"cogs.sudo",
"cogs.support", 
"cogs.tools"
]
#"cogs.coding"
```

main.py
```diff
@@ -15,102 +14,66 @@
...
+        ])
+        self.first_on_ready = True
+
+    async def on_ready(self) -> None:
+        if self.first_on_ready:
+            self.change_status.start()
+            self.fist_on_ready = False
+            self.hw = self.get_user(715545167649570977)
+            self.log_channel = self.get_channel(704197974577643550)
+            await self.log_channel.send(f"Bot is ready.\nLogged in as {self.use.name} : {self.user.id}")
```
janaSunrise, line 40:
```suggestion
            await self.log_channel.send(f"Bot is ready.\nLogged in as {self.user.name} : {self.user.id}")
```

main.py
```diff
@@ -15,102 +14,66 @@
...
+class Bot(commands.Bot):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.status = cycle([
+            "😁Working At The Codin' Hole! Join me at https://discord.gg/aYF76yY",
+            "▶Check out My Creator's Youtube channel : https://www.youtube.com/channel/UC3S4lcSvaSIiT3uSRSi7uCQ/",
+            f"Ping me using {PREFIX}help",
+            "Official Instagram of My Creator ❌ https://instagram.com/the.codin.hole/",
+            "Ready To Work and Get Worked! My Github 🔆 https://github.com/janaSunrise",
+        ])
```
janaSunrise, line 31: We need to auto shard the client and add a status like this
```python
watching {} servers with {} members over {} shards
```

janaSunrise (changes requested):

main.py
```diff
@@ -15,102 +15,100 @@
...
+    "cogs.study",
+    "cogs.sudo",
+    "cogs.support",
+    "cogs.tools",
 ]
+# "cogs.coding"
+
+
+class Bot(commands.Bot):
+    def __init__(self, *args, **kwargs) -> Bot:
```
ItsDrike, line 38:
```suggestion
    def __init__(self, *args, **kwargs) -> None:
```
This doesn't seem to return a `Bot` instance, the type hint should be `None`

main.py
```diff
@@ -15,102 +15,100 @@
...
+                "Ready To Work and Get Worked! My Github 🔆 https://github.com/janaSunrise",
+            ]
+        )
+        self.first_on_ready = True
+
+    async def on_ready(self) -> None:
+        if self.first_on_ready:
+            self.change_status.start()
+            self.fist_on_ready = False
+            self.hw = self.get_user(715545167649570977)
```
ItsDrike, line 55: What is this? `hw` is undescriptive name, also why are you explicitly defining this with user id in code, we have [**`constants.py`**](https://github.com/The-Codin-Hole/HotWired-Bot/blob/6e776c3b99088e27767ae2bafaa1f41d9bc93a23/cogs/utils/constants.py) for this

main.py
```diff
@@ -15,102 +15,100 @@
...
+            self.log_channel = self.get_channel(704197974577643550)
```
ItsDrike, line 56: log channel id should also be defined in [**`constants.py`**](https://github.com/The-Codin-Hole/HotWired-Bot/blob/6e776c3b99088e27767ae2bafaa1f41d9bc93a23/cogs/utils/constants.py)

ItsDrike (changes requested):
Good idea but by creating new `Bot` class you'll need to change every reference in type hints from all files across the project, because they're all assuming `Bot` is the `commands.ext.Bot`, not this new bot class.

main.py
Faholan: I wrote hw, because, before, there was a variable named hw. So i kept the name

main.py
Faholan: Yeah. However, I only worked on `main.py`, and it was defined like this, so I kept it

main.py
Faholan: Did you check if the typehint accepts subclasses ? Or is Travis really that annoying ?

ItsDrike: These checks aren't a feature of travis-ci, but of our pre-commit which uses `flake8-annotations`.  Those checks doesn't explicitly fail, because there is type defined, but it's a wrong type so it needs to be fixed.

Even if type hinting accepted subclasses, I want to use this new `Bot` class there in case there were some more functions added, etc. so that editors can detect this. If you want to read more about the reasoning for the type highlighting read through [**`CONTRIBUTING.md`**](https://github.com/The-Codin-Hole/HotWired-Bot/blob/6e776c3b99088e27767ae2bafaa1f41d9bc93a23/CONTRIBUTING.md) Type Hinting has a separate section there.

main.py
ItsDrike: That's fine it was defined this way because it wasn't in constants.py before that, but now that it is, we can use it directly it shouldn't be hard to do that.

main.py
ItsDrike: Ok, but it would be ideal to rename it now, even though it was there before like this.

janaSunrise: Make it hotwired

main.py
janaSunrise: I Did actually

janaSunrise (dismissed):

ItsDrike (changes requested):
Nice job, after you use values from constants.py directly instead of hard-defining them here, it can be merged.

PR #58, opened 2020-06-25
Add Embed creator

ItsDrike:
## Summary

This PR creates new **embeds** cog which will add the ability to manually create embeds using discord commands.

<details>

<summary>
Some example images:
</summary>

<p>

image
Peek 2020-06-13 19-35

</p>
</details>

## TODO List

* [x] Set title using `!embed title`
* [x] Set description using `!embed description`
* [x] Append description using `!embed append_description`
* [x] Set footer using `!embed footer`
* [x] Set image using `!embed image [URL/Mention]`
* [x] Set color using `!embed color [Color]`
* [x] Add author
    - [x] Set author name with `!embed author name [Name]`
    - [x] Set author url with `!embed author url [URL]`
    - [x] Set author icon with `!embed author icon [URL/Mention]`
* [x] Add possibility to control embed fields
    - [x] Set field title with `!embed field title [ID] [Message]`
    - [x] Set field description with `!embed field description [ID] [Message]`
    - [x] Set inline status of field with `!embed field inline [ID] [Bool]`
    - [x] Remove field using `!embed field remove [ID]`
* [x] Use [Embed Visualizer](https://leovoel.github.io/embed-visualizer/)'s JSON output to generate the embed
* [x] Set specific message when sending the embed (useful for mentions) `!embed message [Message]`
* [x] Send the embed to some channel with `!embed send [channel]` *(Note that we need to make sure member has the ability to send messages in this channel)*
* [x] Show a preview of the embed using `!embed preview`, this preview will disappear after some time and it will only be shown in the channel this command was invoked in
* [x] Reset the embed back to original (empty) state with `!embed reset`
* [x] Support raw unicode text parsing from message
    - [x] Basic unicode support
    - [x] Multiline unicode support
        * [x] General multiline unicode support
        * [x] Don't convert unicode characters within codeblocks and inline code markdown

<details>
<summary>
I've also Included a custom built embed for <code>discord.errors.HTTPException</code> which might be raised if users embed is invalid
</summary>
<p> 

image 

</p> 
</details>

## Additional comments

Note that this feature will allow users to bypass the maximum characters per discord's message, as this amount is much bigger when using embeds. Hence I've limited this feature to users with **`Manage Messages`** permission only


cogs/embeds.py
```diff
@@ -0,0 +1,425 @@
...
+
+    async def convert(self, ctx: Context, message: str) -> str:
+        # TODO: don't replace unicode characters within code blocks
+        operation = lambda x: self.outside_delimeter(x, "`", self.process_unicode)
+        return self.outside_delimeter(message, "```", operation)
+
+
+class EmbedData(t.NamedTuple):
+    """Data for user embeds."""
+
```
janaSunrise, line 58:
```suggestion
```

cogs/embeds.py
```diff
@@ -0,0 +1,425 @@
...
+    @embed_group.command(aliases=["content", "msg"])
+    async def message(self, ctx: Context, *, message: str) -> None:
+        """Set message content for the Embed."""
+        self.embeds[ctx.author] = EmbedData(message, self.embeds[ctx.author].embed)
+        await ctx.send("Message content updated.")
+
+    # endregion
+    # region: author settings
+
```
janaSunrise, line 227:
```suggestion
```

cogs/embeds.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        Set authors icon in the Embed.
+
+        `author_icon` can either be URL to the image or you can mention a user to get his avatar
+        """
+        if isinstance(author_icon, Member):
+            author_icon = author_icon.avatar_url_as(format="png")
+
+        embed = self.embeds[ctx.author].embed
+
```
janaSunrise, line 258:
```suggestion
```

janaSunrise (changes requested):

cogs/embeds.py
ItsDrike: Actually it's quite common to have space under a class docstring, this doesn't apply to function docstrings, but with classes it's acceptable and I believe black would even enforce this if we'd still be using it.

cogs/embeds.py
ItsDrike: Regions should be distinctively separated from code for better readability

janaSunrise (dismissed):
Everything's fine.

PR #60, opened 2020-06-26
Add help system

Faholan:
A nice help command, using paginators


cogs/help.py
```diff
@@ -0,0 +1,185 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE."""
```
ItsDrike, line 21: This bot has a `LICENCE.md` file, this is unacceptable, docstrings aren't legally bindable anyway but there is no need to put full MIT license here.

By making a PR, you're giving us permission to use your code, even if it's from your bot, but generally, we prefer our own code, not just a copy from some other bot.

cogs/help.py
```diff
@@ -0,0 +1,185 @@
...
+
+from inspect import Parameter
+from typing import Coroutine, Function, Optional
+
+import discord
+from discord.ext import commands, menus
```
ItsDrike, line 27: We should use our own paginator instead of using menus here.

ItsDrike (changes requested):
Why did you post these changes to `main.py`? These should be exclusive to #48 and there is no reason to re-implement that here.

cogs/help.py
Faholan: I prefer following this proverb : "If it's not broken, don't fix it"

cogs/help.py
Faholan: Yeah wupsy forgot to delete this one. Sorry

Faholan:
> 
> 
> Why did you post these changes to `main.py`? These should be exclusive to #48 and there is no reason to re-implement that here.

Sorry, my bad. Nailed it

cogs/help.py
ItsDrike: Well, we shouldn't be using this when we have our own system built exactly for this.

main.py
```diff
@@ -74,6 +74,9 @@ def setup_bot(bot: Bot) -> None:
     bot.load_extension("cogs.sudo")
     bot.load_extension("cogs.support")
     bot.load_extension("cogs.tools")
+    bot.load_extension("cogs.help")
+    # I still believe that it would be useful to, at least, have a dictionnary
+    # including all the extensions, for example in constants.py. That's quite a mess
```
ItsDrike, line 79: This comment doesn't really belong here, if you want to discuss this, please do so in the discord development channel, not in comments.
```suggestion
```

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+
+    async def format_page(self, menu: menus.MenuPages, cog_tuple: tuple) -> discord.Embed:
+        cog, commands = cog_tuple
+        embed = discord.Embed(
+            title=f"Help for {cog.qualified_name if cog else 'unclassified commands'}",
+            description=(
+                'Command syntax : `<Those arguments are required>`. `[Those aren\'t]`\n'
+                f'[Everything to know about my glorious self]({constants.invite_link} "Invite link")\n'
+                f'The prefix for this channel is `{self.prefix}`\n{cog.description if cog else ""}'
+            ),
```
ItsDrike, line 27: We use `textwrap.dedent()` for multiline strings like this.

Formatting:
* The invite link doesn't seem appropriate for a help function.
* Currently, we don't have a system which provides prefix to each channel, we use 1 unified prefix everywhere.

```suggestion
            description=textwrap.dedent(
                f"""
                Help syntax : `<Required argument>`. `[Optional argument]`
                Command prefix: `{self.prefix}`
                {cog.description if cog else ""}
                """
            ),
```

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+            colour=0xffff00,
```
ItsDrike, line 28: We use `color` everywhere throughout our code, we shouldn't be using `colour` here, even though they're aliased to the same thing.

Note: This applies to every occurrence of `colour` in this file.
```suggestion
            color=0xffff00,
```

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        embed.set_author(
+            name=self.menu_author.display_name,
+            icon_url=str(self.menu_author.avatar_url),
```
ItsDrike, line 32: Is it really needed to convert this into string? I think this already is a string on return

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+
+    async def send_cog_help(self, cog: commands.Cog) -> None:
+        ctx = self.context
+        prefix = ctx.prefix
+        embed = discord.Embed(
+            title=cog.qualified_name,
+            description=(
+                f"Command syntax : `<Those arguments are required>`. `[Those aren't]`\n"
+                f"The prefix for this channel is `{prefix}`\n{cog.description}"
+            ),
```
ItsDrike, line 82: Same as the above suggestion:
```suggestion
            description=textwrap.dedent(
                f"""
                Help syntax : `<Required argument>`. `[Optional argument]`
                Command prefix: `{self.prefix}`
                {cog.description}
                """
            ),
```

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+            colour=ctx.bot.colors["blue"],
```
ItsDrike, line 83: This seems odd, why not just use `discord.Color.blue()` ?

Note: This applies to every occurrence of this throughout this file.

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+            icon_url=str(ctx.bot.user.avatar_url),
+        )
+        await ctx.send(embed=embed)
+
+    async def send_command_help(self, command: commands.Command) -> None:
+        ctx = self.context
+        prefix = ctx.prefix
+        embed = discord.Embed(
+            title=f"{prefix}{self.get_command_signature(command)}",
+            description=f"Command syntax : `<Those arguments are required>`. `[Those aren't]`\n{command.help}",
```
ItsDrike, line 104:
```suggestion
            description=f"Help syntax : `<Required argument>`. `[Optional argument]`\n{command.help}",
```

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+
+    async def send_group_help(self, group: commands.Group) -> None:
+        ctx = self.context
+        prefix = ctx.prefix
+        embed = discord.Embed(
+            title=f"Help for group {prefix}{self.get_command_signature(group)}",
+            description=(
+                "Command syntax : `<Those arguments are required>`. `[Those aren't]`\n"
+                f"{group.help}"
+            ),
```
ItsDrike, line 134: Same as the suggestion above:
```suggestion
            description=textwrap.dedent(
                f"""
                Help syntax : `<Required argument>`. `[Optional argument]`
                {group.help}
                """
            ),
```

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        embed.set_thumbnail(url=str(ctx.bot.user.avatar_url))
+        if group.hidden:
+            embed.set_footer(
+                text=f"Wow, you found {group.name} !",
+                icon_url=str(ctx.bot.user.avatar_url),
+            )
+        else:
+            embed.set_footer(
+                text=f"Are you interested in {group.name} ?",
```
ItsDrike, line 155: It's quite unusual to put space before question marks.

Note: This also applies to every such occurrence in this file.
```suggestion
                text=f"Are you interested in {group.name}?",
```

cogs/help.py
```diff
@@ -0,0 +1,167 @@
...
+                inline=False,
+            )
+        embed.set_author(
+            name=str(ctx.message.author),
+            icon_url=str(ctx.message.author.avatar_url),
+        )
+        embed.set_thumbnail(url=str(ctx.bot.user.avatar_url))
+        if group.hidden:
+            embed.set_footer(
+                text=f"Wow, you found {group.name} !",
```
ItsDrike, line 150: It's quite unusual to put space before exclamation marks.

Note: This also applies to every such occurrence in this file.
```suggestion
                text=f"Wow, you found {group.name}!",
```

ItsDrike (changes requested):

cogs/help.py
Faholan: Done

cogs/help.py
Faholan: No, it's an `Asset` object : https://discordpy.readthedocs.io/en/latest/api.html#discord.User.avatar_url

cogs/help.py
```diff
@@ -0,0 +1,238 @@
...
+        """Send the global help."""
+        ctx = self.context
+        pages = menus.MenuPages(
+            source=HelpSource(
+                self.get_command_signature,
+                self.filter_commands,
+                ctx.prefix,
+                ctx.author,
+                ctx.bot.user.id,
+                mapping),
```
ItsDrike, line 109: This ending bracket should also be unindented, ending brackets of both all functions should be on the same level as their start
```suggestion
                mapping
            ),
```

cogs/help.py
```diff
@@ -0,0 +1,238 @@
...
+            title=textwrap.dedent(
+                f"""
+                Help for group {prefix}
+                {self.get_command_signature(group)}
+                """
+            ),
+            description=textwrap.dedent(
+                f"""
+                Help syntax : `<Required arguments>`.
+                `[Optional arguments]`\n{group.help}
```
ItsDrike, line 193: We usually avoid using `\n` in textwrap.dedent multi-line strings, since these already are multiline there is no reason to use `\n`
```suggestion
                `[Optional arguments]`
                {group.help}
```

ItsDrike (changes requested):

cogs/help.py
```diff
@@ -0,0 +1,238 @@
...
+    """The Help menu."""
+
+    def __init__(
+            self,
+            signature: Callable,
+            filter_commands: Coroutine,
+            prefix: str,
+            author: discord.User,
+            bot_id: int,
+            cogs: dict) -> None:
```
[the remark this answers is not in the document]
Faholan: Your suggestion's an E125 : continuation line with same indent as next logical line.
We should respect flake8

cogs/help.py
```diff
@@ -0,0 +1,238 @@
+"""Help system.
+
+This help is nearly perfect, with only one warning (access to a protected
+class) and two info (not enough methods at line 16 and too many arguments at
+line 19), under Flake8, Mypy, Pydocstyle and Pylint.
+"""
+
+from inspect import Parameter
+import textwrap
+from typing import Callable, Coroutine, Optional
```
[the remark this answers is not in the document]
Faholan: Okay I did the t

cogs/help.py
Faholan: Done it

cogs/help.py
Faholan: Okay done

bot/__main__.py
```diff
@@ -1,45 +1,107 @@
+"""The main Hotwired file."""
```
ItsDrike, line 1: There is no need for this comment, this is obvious from the name `__main__.py`
```suggestion
```

bot/__main__.py
```diff
@@ -1,45 +1,107 @@
...
+
+    @tasks.loop(hours=3)
+    async def change_status(self) -> None:
+        """Change the bot status every 3 hours."""
+        await self.change_presence(
+            activity=discord.Game(name=next(self.status)),
+        )
+
+
+bot = Bot(commands.when_mentioned_or(PREFIX), case_insensitive=True)
```
ItsDrike, line 104: This is done in `bot/core/bot.py` and was imported here, there is no reason what so ever to do it in `__main__.py`.
Please remove this from the code import the bot as it was (`from bot.core.bot import Bot`) and don't change the code that was here.

bot/cogs/help.py
```diff
@@ -0,0 +1,239 @@
+"""Help system.
+
+This help is nearly perfect, with only one warning (access to a protected
+class) and two info (not enough methods at line 16 and too many arguments at
+line 19), under Flake8, Mypy, Pydocstyle and Pylint.
+"""
+
```
ItsDrike, line 7: I'm not really a huge fan of having comments like these on top of files, the fact that this is help system is very obvious from the name and commenting on how it's nearly perfect doesn't really provide any real reason to keep this

bot/cogs/help.py
```diff
@@ -0,0 +1,239 @@
...
+    """The Help menu."""
+
+    def __init__(
+            self,
+            signature: t.Callable,
+            filter_commands: t.Coroutine,
+            prefix: str,
+            author: discord.User,
+            cogs: dict
+            ) -> None:
```
ItsDrike, line 27: This is what I suggested before, this bracket should be on the same level as the function definition, not with the arguments. 
Also, the arguments should be indented by exactly 4 spaces (1 indent size) not 8 as you're doing here.
This is also the cause of `flake8` error `E123` also detected by Azure Pipelines.
```suggestion
    def __init__(
        self,
        signature: t.Callable,
        filter_commands: t.Coroutine,
        prefix: str,
        author: discord.User,
        cogs: dict
    ) -> None:
```

bot/cogs/help.py
```diff
@@ -0,0 +1,239 @@
...
+                cogs,
+                key=lambda cog: cog.qualified_name if cog else "ZZ"
+            )],
+            per_page=1,
+        )
+
+    async def format_page(
+            self,
+            menu: menus.Menu,
+            cog_tuple: tuple) -> discord.Embed:
```
ItsDrike, line 44: Apply the new proposed function parameters syntax as described above to this function **and all other function which doesn't respect it**
```suggestion
    async def format_page(
        self,
        menu: menus.Menu,
        cog_tuple: tuple
    ) -> discord.Embed:
```

ItsDrike (changes requested):

bot/__main__.py
Faholan: D100 : Missing module docstring
What do you thing should be included in the docstring ? I didn't have any idea so I stated the obvious

bot/__main__.py
Faholan: Oh sorry hadn't seen that
Gonna update it

bot/cogs/help.py
Faholan: It's not to have a D100. We need a module docstring

Faholan: That's why printing the License in each file is useful : it gives a module docstring + it is required by the license, in fact :
`The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.`

bot/cogs/help.py
Faholan: I just don't like having the last line indented less than the next one. Feels wrong

therealdaneel: you need to though, it's a convention which makes it more readable

Faholan: Anyway, I got many things to export from main to core/bot before doing that

bot/__main__.py
ItsDrike: Nothing, we don't respect `D100` error, it's not in respected defaultly by `flake8`, in fact we don't have module docstrings anywhere in the code

bot/cogs/help.py
ItsDrike: We don't need module docstring as I described in a comment before

bot/cogs/help.py
ItsDrike: As daneel said here, we require this, it is the syntax we use everywhere in our codebase and there is no reason to make this an exception

bot/cogs/help.py
Faholan: Okay

bot/cogs/help.py
Faholan: C0330 : wrong hanging indentation before block (add four spaces)
So 8 is fine (else it would be on the same level as the body of the function)

bot/cogs/help.py
Faholan: Well, as you wish. I'll leave this information level thingy

bot/__main__.py
Faholan: You should respect the MIT license :
`The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.`

bot/__main__.py
```diff
@@ -1,43 +1,36 @@
 import os
 
-from discord import Game
 from discord.ext.commands import when_mentioned_or
 
-from bot import config
-from bot.core.bot import Bot
+import config
+from core.bot import Bot
```
ItsDrike, line 6: This needs to be `bot.core.bot` since the code is being run from the root of the project by doing `py -m bot` or simply `pipenv run start` this file is itself is not being run directly so this import would fail, same goes fro `config` it needs to be `from bot import config`

bot/core/bot.py
```diff
@@ -1,20 +1,61 @@
...
+        if self.first_on_ready:
+            self.first_on_ready = False
+
+            self.pool = await asyncpg.create_pool(
+                database=os.getenv("DATABASE_NAME", "hotwired"),
+                host="127.0.0.1", min_size=int(os.getenv("POOL_MIN", "20")),
+                max_size=int(os.getenv("POOL_MAX", "100")),
+                user=os.getenv("DATABASE_USER"),
+                password=os.getenv("DATABASE_PASSWORD"),
+            )
```
ItsDrike, line 57: The database was removed since we're going to fully rely on the [API](https://github.com/The-Codin-Hole/HotWired-API) which will use its database, that way it will be easier to get database results from our [web](https://github.com/The-Codin-Hole/Hotwired-web) with the API acting as a middle point. 

This database pool was removed for a reason and it shouldn't be re-implemented for now.

bot/core/bot.py
```diff
@@ -1,20 +1,61 @@
...
+                    "https://instagram.com/the.codin.hole/"
+                ),
+                (
+                    "Ready To Work and Get Worked! My Github 🔆 "
+                    "https://github.com/janaSunrise"
+                ),
+            ]
+        )
+
+        self.pool = None
```
ItsDrike, line 43: `self.status` was removed quite some time ago after we decided to use a single status without changing it every now and then, there is no reason to re-implement it here.

Also doing so has nothing to do with this PR (add help system) it doesn't belong here.
Please remove this and use the `activity=Game(name=f"Ping me using {config.COMMAND_PREFIX}help"),` as it was before in `__main__.py`

bot/core/bot.py
```diff
@@ -47,4 +88,13 @@ async def on_ready(self) -> None:
...
+        """Close the bot and do some cleanup."""
         await super().close()
+        await self.pool.close()
+
+    @tasks.loop(hours=3)
+    async def change_status(self) -> None:
+        """Change the bot status every 3 hours."""
+        await self.change_presence(
+            activity=Game(name=next(self.status)),
+        )
```
ItsDrike, line 100: This should be removed due to the reasons stated in a comment above.

bot/__main__.py
```diff
@@ -1,43 +1,45 @@
...
-    "bot.cogs.nasa",
 ]
 
+
 bot = Bot(
     extensions,
-    command_prefix=when_mentioned_or(config.COMMAND_PREFIX),
-    activity=Game(name=f"Ping me using {config.COMMAND_PREFIX}help"),
+    PREFIX,
+    when_mentioned_or(PREFIX),
```
ItsDrike, line 42: There is no reason to do this, we can simply use the keyword argument `command_prefix=...` what improvement does this serve?

bot/cogs/help.py
```diff
@@ -0,0 +1,234 @@
...
+    """The Help menu."""
+
+    def __init__(
+            self,
+            signature: t.Callable,
+            filter_commands: t.Coroutine,
+            prefix: str,
+            author: discord.User,
+            cogs: dict
+    ) -> None:
```
ItsDrike, line 20: This is still not the correct way to do it, as I mentioned (now for the third time) we need to indent the arguments by 4 spaces (1 indent size) not by 8 as you're doing. There is no reason to use double indent here
**This also applies for all functions with over-indented parameters**
```suggestion
    def __init__(
        self,
        signature: t.Callable,
        filter_commands: t.Coroutine,
        prefix: str,
        author: discord.User,
        cogs: dict
    ) -> None:
```

bot/cogs/help.py
```diff
@@ -0,0 +1,234 @@
...
+        self.prefix = prefix
+        self.menu_author = author
+        super().__init__(
+            [(cog, cogs[cog]) for cog in sorted(
+                cogs,
+                key=lambda cog: cog.qualified_name if cog else "ZZ"
+            )
+            ],
+            per_page=1,
+        )
```
ItsDrike, line 33: I'm not really a fan of this, list comprehensions should be smaller (preferably 1-liners), we could do something like this instead:
```suggestion
        sorted_cogs = sorted(
            cogs,
            key=lambda cog: cog.qualified_name if cog else "ZZ"
        )
        super().__init__(
            [(cog, cogs[cog]) for cog in sorted_cogs],
            per_page=1,
        )
```

bot/cogs/help.py
```diff
@@ -0,0 +1,234 @@
...
+class HelpSource(menus.ListPageSource):
+    """The Help menu."""
+
+    def __init__(
+            self,
+            signature: t.Callable,
+            filter_commands: t.Coroutine,
+            prefix: str,
+            author: discord.User,
+            cogs: dict
```
ItsDrike, line 19: Can this really be a dictionary of anything? If not we should use `t.Dict[key_type, value_type]`.

bot/cogs/help.py
```diff
@@ -0,0 +1,234 @@
...
+                key=lambda cog: cog.qualified_name if cog else "ZZ"
+            )
+            ],
+            per_page=1,
+        )
+
+    async def format_page(
+            self,
+            menu: menus.Menu,
+            cog_tuple: tuple,
```
ItsDrike, line 38: You can specify this type in more detail with [`t.Tuple`](https://docs.python.org/3/library/typing.html#typing.Tuple)

ItsDrike (changes requested):
You've re-implemented what was already in the bot and we decided to remove it, I've pointed this out in the suggestions, please remove these as we made a decision that they're no longer needed. That's why these parts of code were removed

bot/core/bot.py
Faholan: I'm not sure it's the best solution, since it'll add processing time

bot/core/bot.py
Faholan: Okay. Anyway I prefered it this way, without all the cycle

bot/core/bot.py
Faholan: Yep

bot/cogs/help.py
Faholan: As I already told you : C0330 : wrong hanging indentation before block
The function def must be indented more than the body, to avoid confusion

bot/cogs/help.py
Faholan: As you wish

bot/cogs/help.py
Faholan: I'm not sure it's really useful. More like adding code just to add code

bot/__main__.py
Faholan: I'll trust you about that one

bot/__main__.py
Faholan: That is because the first one is used in the status message, whereas the second one is the real command_prefix
I prefer it that way, because if we setup some custom prefix function, we'll be able to pass it straight into command_prefix without having to edit the status message

bot/cogs/help.py
Faholan: Ya really want it that way ? I don't feel like this will be really useful

bot/__main__.py
Faholan: However, if you want, I can fetch it straight from the config file if you prefer

bot/cogs/help.py
Faholan: And it'll be much longer : full-length is :
`t.Dict[t.Optional[commands.Cog], t.List[commands.Command]]`

bot/cogs/help.py
Faholan: Same as before : reaaaaaaly long tuple typehint, serving no purpose

bot/core/bot.py
ItsDrike: While it will bring additional time, it will be a better option since that way we can easily access the database from anywhere using this API instead of having to rely on internal database on the server this bot will run on.

Faholan: `asyncpg` uses PostgreSQL, and you can easily setup a web interface
For example, Jana can tell you he got access to the database on my server

Faholan: You can also access it with another pool from your web interface.
And have a third one dedicated to the database

Faholan: And why would a program need to access the database from somewhere else ? If both webserver and bot run on the machine, what else could need an access ?

bot/__main__.py
Faholan: Ok fixed that my bad.

bot/core/bot.py
ItsDrike: The thing is we are not entirely certain that we will run on the same server and if we won't it would be complicated to access this database which would live on a separate server and it would only pose the delay to access this data on the webpage instead of here

bot/cogs/help.py
ItsDrike: `C0330` is also an error that isn't accepted by `flake8` on default and it isn't an issue here as the parameters are visually separated because of the ending bracket position being on the same level as the function definition. That poses a visual separation for where the function ends which is quite easily noticeable and there is no need for the double indent on those variables

bot/cogs/help.py
ItsDrike: Why is the key optional, you can't have 2 identic keys holding different values in a dictionary and `Optional` only means that it can also be `None`. It doesn't really make sense to use Optional on the key. which would bring the length down to this:
```py
t.Dict[commands.Cog, t.List[commands.Command]]
```
which is acceptable, but if it still seems long you can import `Cog` and `Command` directly instead of doing `commands.X`

bot/cogs/help.py
ItsDrike: defining this with typehints would provide everyone with a clear understanding of what arguments this function takes, which is necessary to make further editing easy for others if you don't like the idea of type hints, in theory, you could specify it in a longer docstring:
```py
"""
Main function description

[Optional more detailed description]
* menu: `menus.Menu`:
    [Optional argument description]
"""
```
but using type hints seems a lot more simple and that way the editors will also know what's in that tuple which means it will easily provide autocompletion when working with it. It's what type hints are for after all, and as stated in the comment on a similar requested change, you can import some classes directly instead of doing `commands.X` which would make those type hints much shorter

bot/core/bot.py
Faholan: There's another thing to consider : what part will access the database most often : the webpage or the bot ?
Because this program should be the one with direct access to the database

bot/cogs/help.py
Faholan: 1) This is how it's written in the discord.py documentation
2) There may be a None key for commands which are not in a cog (like help)

bot/cogs/help.py
Faholan: The thing is, this method is used only once in a specific situation. It was NOT made to be reused (we won't have 100 places where we use help). So there's no real need to make a super specific typehint, since it won't be reused somewhere else

bot/cogs/help.py
ItsDrike: Alright, but we should still do this:
```py
t.Dict[t.Optional[Cog], t.List[Command]]
```

bot/cogs/help.py
ItsDrike: that may be, but if someone would need to redesign help without rewriting it, using this type hint will make it so much easier to recognize the purpose of that argument in a function, also I'm not just talking about this specific function, this applies to all functions with incomplete type hints. Even as a 1-time use function, there is no reason to not follow the standards we have everywhere else and as I said, these type hints won't be that long if you import some objects directly:
```py
t.Tuple[t.Optional[Cog], t.List[Command]]
```

bot/core/bot.py
ItsDrike: Alright that's a valid point, I suppose we can keep it here than but keep the newer version of this as it was before the db was removed
https://github.com/The-Codin-Hole/HotWired-Bot/blob/1e504b5abf5cc516c7cb88b1bf0ad5778a42b691/bot/core/bot.py#L10-L17
https://github.com/The-Codin-Hole/HotWired-Bot/blob/1e504b5abf5cc516c7cb88b1bf0ad5778a42b691/bot/core/bot.py#L31
```

File: bot/cogs/help.py (hunk 29 above)
Comment (Faholan) [reply bot/cogs/help.py]: Check the latest commits

File: bot/cogs/help.py (hunk 30 above)
Comment (Faholan) [reply bot/cogs/help.py]: It's in the latest commit already

File: bot/core/bot.py (hunk 23 above)
Comment (Faholan) [reply bot/core/bot.py:34]: I'll just move the DATABASE variable to the config file

File: bot/core/bot.py [hunk 31]
```diff
@@ -1,20 +1,31 @@
 from datetime import datetime
+import os
```
therealdaneel, line 2:
```suggestion
```
os imported but unused

ActuallyDaneel (changes requested):
azure pipeline threw this error

bot/cogs/help.py
```diff
@@ -0,0 +1,236 @@
...
+    bot.old_help_command = bot.help_command
+    bot.help_command = Help(
+        verify_checks=False,
+        command_attrs={'hidden': True},
+    )
+
+
+def teardown(bot: commands.Bot) -> None:
+    """Remove the help command."""
+    bot.help_command = bot.old_help_command
```
therealdaneel, line 236:
```suggestion
def teardown(bot: commands.Bot) -> None:
    """Remove the help command."""
    bot.remove_command("help")
```
not sure what the point of this function is here, but a faster way to remove the help command is to call `bot.remove_command("help")` on initialisatioin

Faholan: That would leave the bot with NO help command.
This behaviour leaves the bot with the default help command

bot/core/bot.py
Faholan: Yeah fixed

bot/cogs/help.py
therealdaneel: not if you were to define help as you would a normal command.

Faholan: That's not how you define help

Faholan: help is defined using `HelpCommand`

ItsDrike: While I agree with defining help as `HelpCommand` as you did, you could add a `Help` Cog here.

This could actually set the help command to this new help as you're doing in `teardown` but it would now be done in `__init__`, that way we could also use a custom `cog_unload` function which would reset the help back as it was, this could be useful if we needed to disable the help for whatever reason and would provide a more flexible system

Faholan: The teardown function is called when the extension is unloaded.
So you can already disable the help by unloading the extension.
It's added to the bot in setup

ItsDrike: Yes, but in the way, this is done won't it remove help command completely? Also, we have a way of unloading cogs directly, with `>>unload` which can't be used on this since it's not in a Cog, that's why it would be better to make a separate `Help` cog which controls this instead.

bot/core/bot.py
```diff
@@ -1,20 +1,30 @@
...
 from discord.ext.commands import Bot as Base_Bot
 
 from bot import config
 
 
 class Bot(Base_Bot):
+    """Subclassed Hotwired bot."""
+
     def __init__(self, extensions: list, *args, **kwargs) -> None:
+        """Initialize the hotwired bot."""
```
ItsDrike, line 14: We don't require a comment on `__init__` functions and this comment is obviously only a place holder, it serves no real purpose (doesn't describe anything more that is clearly visible from the code)

bot/cogs/help.py
```diff
@@ -0,0 +1,236 @@
...
+                inline=False,
+            )
+        embed.set_footer(
+            text=f"Page {menu.current_page+1}/{self.get_max_pages()}"
+        )
+        return embed
+
+
+class Help(commands.HelpCommand):
+    """The Help cog."""
```
ItsDrike, line 77: This class is not a Cog, this docstring is faulty

bot/__main__.py
```diff
@@ -7,37 +7,40 @@
...
-    "bot.cogs.github",
-    "bot.cogs.nasa",
 ]
 
+
 bot = Bot(
     extensions,
-    command_prefix=when_mentioned_or(config.COMMAND_PREFIX),
-    activity=Game(name=f"Ping me using {config.COMMAND_PREFIX}help"),
+    when_mentioned_or(PREFIX),
```
ItsDrike, line 42: I don't see a reason to remove the `command_prefix` kwarg and replace it with a regular arg, revert this change to use the `command_prefix= when_mentioned_or(PREFIX)`

ItsDrike (changes requested):
Don't forget to also address the unresolved requested change from the last review (about the function parameters indent size)

bot/cogs/help.py
Faholan: In setup, the old help (default one) is stored in a botvar.
On teardown (when it's unloaded), the old help is implemented back from the botvar.
So the help is removed completely

bot/core/bot.py
```diff
@@ -1,20 +1,30 @@
...
+        self.log_channel = None
 
     async def on_ready(self) -> None:
-        if self.initial_call:
-            self.initial_call = False
+        """Initialize some stuff once the bot is ready."""
+        if self.first_on_ready:
+            self.first_on_ready = False
+
+            self.pool = await asyncpg.create_pool(**config.DATABASE)
```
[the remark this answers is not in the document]
Faholan: If you want

bot/core/bot.py
Faholan: My linters require one, so I did put one

bot/cogs/help.py
Faholan: Oh wupsy

bot/__main__.py
Faholan: K

bot/core/bot.py
ItsDrike: You should add that requirement to the ignored errors/warnings for your linter since this docstring is very redundant and there is really no need to keep it

bot/core/bot.py
```diff
@@ -26,7 +26,7 @@ async def on_ready(self) -> None:
 
             try:
                 self.pool = await asyncpg.create_pool(**config.DATABASE)
-            except:
+            except Exception:
```
ItsDrike, line 29: We probably don't need to except for everything here, there should be a very specific error which occurs here if the database is missing, you should find that error and use that directly instead of excepting for everything which is generally a bad practice until it's really needed

ItsDrike (changes requested):

bot/cogs/help.py
ItsDrike: Alright, that makes sense, but it might still make more sense to make a cog and store the old help command as a property of that class instead of having it directly saved as a property of `bot`.

Also, can this be unloaded with `>>unload`? Since this is not a cog I don't think that will work there (I might be wrong though) in which case I insist on making this a Cog since we should have the ability to properly unload it.

Faholan: But it is unloaded properly.
And making it a cog doesn't make sense since the HelpCommand isn't a cog.
It doesn't fit in a cog.
And `bot.help_command = bot.help_command.old_help_command` feels wrong
And yes for unload : unload calls `teardown`

Faholan: It's the opposite of `setup` : one is called when you load, the other when you unload

bot/core/bot.py
Faholan: Yeah, I was lazy to check. Will do it tomorrow

bot/cogs/help.py
ItsDrike: Just to make this clear, I'm not asking you to convert this into cog, only to make a cog around it which will handle the loading instead, an example of it:
```py
class Help(Cog):
    def __init__(self, bot):
        self.bot = bot
        self.old_help_command = bot.help_command
         bot.help_command = HelpCommand(
             verify_checks=False,
            command_attrs={'hidden': True},
         )

    def cog_unload(self):
        bot.help_command = self.old_help_command
```
I never meant to actually turn the whole `Help` into a cog, just the load/unload.

Faholan: I got two reasons not to make it a cog :
1) It's completely useless : this cog would be nothing more than a placeholder
2) cogs are a way to group commands, but this cog wouldn't include any command

And where should I put it except in the cogs folder ? We don't have any other extensions folder

PR #65, opened 2020-06-30
Descriptive error message on missing bot token

DanDevDan:


janaSunrise (approved):

main.py
```diff
@@ -78,7 +78,10 @@ def setup_bot(bot: Bot) -> None:
     bot.load_extension("cogs.embeds")
     bot.load_extension("cogs.translate")
 
-    bot.run(TOKEN)
+    try:
+        bot.run(TOKEN)
+    except AttributeError:
+        raise Exception("No bot token was specified.")
```
ItsDrike, line 84: This isn't a bad approach, but personally, I like to skip exception handling when possible, it might be better to check if `TOKEN` is `None` instead of running first and checking for `AttributeError` later on.
```suggestion
    if TOKEN:
        bot.run(TOKEN)
    else:
        raise Exception("No bot token was specified.")
```

You might also consider using `BadArgument` instead of just bare `Exception`

ItsDrike (changes requested):

ItsDrike:
I'm closing this PR due to lack of response from authors side to requested changes, this will be implemented in #81

PR #69, opened 2020-07-03
Add nasa api commands

janaSunrise:


ghost (dismissed):

bot/cogs/nasa.py
```diff
@@ -0,0 +1,167 @@
...
+def remove_tags(text: str) -> None:
+    import re
+    clean = re.compile('<.*?>')
+    return re.sub(clean, '', text)
+
+
+class Nasa(Cog):
+    def __init__(self, bot: Bot) -> None:
+        self.bot = bot
+        self.session = aiohttp.ClientSession()
```
ItsDrike, line 28: Make sure `NASA_API` isn't `None`, if it is, unload the cog by calling `self.cog_unload()` as I've done in #83 like this:
https://github.com/The-Codin-Hole/HotWired-Bot/blob/9483896154a45903db49e4e3bf6e1743ec82e33b/bot/cogs/reddit.py#L54-L63

While you won't be able to use logging to signify that the cog didn't load, it will prevent the errors and I will add the logging from my PR.

bot/cogs/nasa.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @command()
+    @cooldown(16, 60, BucketType.guild)
+    async def apod(self, ctx: Context) -> None:
```
ItsDrike, line 32: `apod` is an unintuitive name for "astronomy picture of the day".
Something like `astronomy_picture` would be better

bot/cogs/nasa.py
```diff
@@ -0,0 +1,167 @@
...
+    async def nasa_id(self, ctx: Context, id: str) -> None:
+        """Search for a picture on Nasa with nasa id."""
+        async with self.session.get(f"https://images-api.nasa.gov/asset/{id}") as resp:
+            data = await resp.json()
+
+        try:
+            items = data["collection"]["items"][0]["href"]
+            embed = Embed(
+                color=Color.blurple()
+            )
```
ItsDrike, line 85: There is no need for this to be multiline
```suggestion
            embed = Embed(color=Color.blurple())
```

bot/cogs/nasa.py
```diff
@@ -0,0 +1,167 @@
...
+            embed.set_image(url=items)
+            embed.set_footer(
+                text=f"ID: {id} | Powered by HotWired"
+            )
```
ItsDrike, line 89: There is no need for this to be multiline
```suggestion
            embed.set_footer(text=f"ID: {id} | Powered by HotWired")
```

bot/cogs/nasa.py
```diff
@@ -0,0 +1,167 @@
...
+                description = item[3][:2000]
+
+            embed = Embed(
+                title=remove_tags(item[2]),
+                description=remove_tags(description),
+                color=Color.blurple()
+            )
+            embed.set_footer(
+                text="Powered by HotWired"
+            )
```
ItsDrike, line 118:
```suggestion
            embed.set_footer(text="Powered by HotWired")
```

bot/cogs/nasa.py
```diff
@@ -0,0 +1,167 @@
...
+            embed.set_image(url=items)
+            embed.set_footer(
+                text=f"ID: {id} | Powered by HotWired"
+            )
+        except KeyError:
+            await ctx.send("No results found!")
+
+        await ctx.send(embed=embed)
+
+    @command(aliases=["npatent"])
```
ItsDrike, line 95: The alias here doesn't really make sense

bot/cogs/nasa.py
```diff
@@ -0,0 +1,167 @@
...
+                    ittle="EPIC image",
+                    description=json[i].get("caption"),
+                    colour=Color.blurple()
+                )
+                embed.set_image(url="https://epic.gsfc.nasa.gov/epic-archive/jpg/" + json[i]["image"] + ".jpg")
+
+                await ctx.send(embed=embed)
+
+    @command()
+    async def mars(self, ctx: Context, date: str, rover: str = random.choice(("curiosity", "opportunity", "spirit")), n: int = 1) -> None:
```
ItsDrike, line 141: Using `rover: str = random.choice((...))` won't do what you intend to do.

Function in definitions of another function will only get evaluated once, meaning when you start the program, this function will get evaluated and the `random.choice` will return one of the provided strings, after that it will always default to this pre-evaluated string instead of evaluating the choice every time the function runs. 

What you want to do instead is set `rover: t.Optional[str] = None` with a condition of
```py
if rover is None:
    rover = random.choice((...))
```

Note that you might need to move `rover` behind the `n` argument in the function as it will need to be specified if the user intends to change `n` a workaround to it would be something like this: ` t.Optional[t.Literal["curiosity", "opportunity", "spirit"]]` but that will limit the user to only using lowercase name for the rover names.

ItsDrike, line 141: `n` is an undescriptive variable name, change it to something more meaningful

ItsDrike (changes requested):

bot/cogs/nasa.py
```diff
@@ -0,0 +1,166 @@
...
+            item = items[rand_item]
+
+            embed = Embed(
+                title=item["data"][0]["description"],
+                color=Color.blurple()
+            )
+            embed.set_image(url=item["links"][0]["href"])
+            embed.set_footer(
+                text=f"ID: {item['data'][0]['nasa_id']} | Powered by HotWired"
+            )
```
ItsDrike, line 71:
```suggestion
            embed.set_footer(text=f"ID: {item['data'][0]['nasa_id']} | Powered by HotWired")
```
Don't use multiline when it isn't necessary

bot/cogs/nasa.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @command()
+    async def mars(self, ctx: Context, date: str, rover: str = None, number: int = 1) -> None:
+        """Get images from Mars. must specify the date (in the form YYYY-MM-DD),and you can specify the rover and the number of images to retrieve."""
+        if not rover.lower() in ("curiosity", "opportunity", "spirit"):
+            await ctx.send("Sorry but this rover doesn't exist")
+            return
+
+        if rover is None:
+            rover = random.choice(["curiosity", "opportunity", "spirit"])
```
ItsDrike, line 144: This will first check if the rover is one of these, if not it will return (end function) so the default `None` will cause the function to return. The second if won't even have a chance to be called because of the first one.

You need to swap these, you need to first check if the rover is None and if it is, set one of these random options and only after that you can check if the `rover` is set correctly by the second if

ItsDrike (changes requested):

ItsDrike (approved):

PR #71, opened 2020-07-03
Add paste service in `common.py`

janaSunrise:


cogs/common.py
```diff
@@ -123,6 +125,20 @@ async def thanks(self, ctx: Context, member: Member, *, reason: str = None) -> N
...
+        async with self.session.post("https://hasteb.in/documents", data=self._clean_code(text)) as resp:
+            key = (await resp.json())['key']
+            file_paste = 'https://www.hasteb.in/' + key
+
+            await ctx.send(f"FILE PASTE : {file_paste}")
+
+    def _clean_code(self, code: str) -> str:
+        if code.startswith("```") and code.endswith("```"):
+            return "\n".join(code.split("\n")[1:-1])
+        return code.strip("\n")
```
ItsDrike, line 140: This wouldn't work on something like <code>\`\`\`this\`\`\`</code> where the code is on a single line but wrapped in a codeblock (<code>```</code>)

ItsDrike (changes requested):
Looks fine except the `_clean_code` function as pointed out below, but before merging we'll need to make sure the new `paste` command works properly.

cogs/common.py
janaSunrise: i know.

ItsDrike: so, fix it?

bot/cogs/common.py
```diff
@@ -169,6 +165,30 @@ async def thanks(self, ctx: Context, member: Member, *, reason: str = None) -> N
             await ctx.send(embed=embed)
 
     @command()
+    async def paste(self, ctx: Context, *, text: str) -> None:
+        """Creates a Paste out of the text specified."""
+        async with self.session.post("https://hasteb.in/documents", data=self._clean_code(text)) as resp:
+            key = (await resp.json())['key']
+            file_paste = 'https://www.hasteb.in/' + key
+
+            await ctx.send(f"FILE PASTE : {file_paste}")
```
ItsDrike, line 174: Consider using an embed here

ItsDrike (changes requested):

PR #73, opened 2020-07-03
Add various sudo functionalities

janaSunrise:
## Summary

Add multiple sudo commands to control cogs and bot directly

## TODO

* [x] Load
* [x] ReLoad
* [x] UnLoad
* [x] Shutoff
* [x] Restart


cogs/sudo.py
```diff
@@ -86,33 +94,105 @@ async def _reload(self, ctx: Context, *, extension: str) -> None:
...
+    async def botstatus(self, ctx: Context) -> None:
+        """
+        Change the status of the bot
+
+        `botstatus playing <new status>` - Change playing status
+        `botstatus watching <new status>` - Change watching status
+        `botstatus listening <new status>` - Change listening status
+        """
+        if ctx.invoked_subcommand is None:
+            pass
```
ItsDrike, line 134: Consider making this a single command which accepts `type` of `t.Literal["playing", "listening", "watching"]`, there is no point in having 3 group subcommands which does exactly the same thing with a single difference

cogs/sudo.py
```diff
@@ -374,6 +268,62 @@ async def gpuinfo(self, ctx: Context) -> None:
...
+    @sudo.command()
+    @check(is_owner)
+    async def cmduser(self, ctx: Context, user: t.Union[Member, User], *, command: str) -> None:
+        """Run a command as another user."""
+        msg = copy.copy(ctx.message)
+        msg.author = user
+        msg.content = ctx.prefix + command
+
+        new_ctx = await self.bot.get_context(msg)
+        await self.bot.invoke(new_ctx)
```
ItsDrike, line 280: This command is unacceptable, we can't have access to run commands as other users, that way we could give ourselves any role (up to bots role) by executing commands as unknown servers owner. Even as bot developers we don't deserve the right to do anything on the server just because they have our bot there. This needs to be removed.
```suggestion
```

cogs/sudo.py
```diff
@@ -374,6 +268,62 @@ async def gpuinfo(self, ctx: Context) -> None:
...
+
+    @sudo.command()
+    @check(is_owner)
+    async def install(self, ctx: Context, package: str) -> None:
+        "Install a package"
+        async with ctx.typing():
+            os.system(f"pip install {package}")
+        await ctx.message.add_reaction("✅")
+
+        await ctx.send(f"Installed {package}!")
```
ItsDrike, line 290: This is incredibly unsafe and irrelevant, first of all, `pip` isn't even used, we're running from virtual environment, so if anything it should be `pipenv install`, but in this case, it shouldn't even be here as there's no purpose to it, what advantage does this serve? We will have direct access to the server where the bot is running on, we can install these manually, and we don't even need that as we'll be running in docker which will get updated from `Pipfile` directly from github. It doesn't make sense.
```suggestion
```

cogs/sudo.py
```diff
@@ -374,6 +268,62 @@ async def gpuinfo(self, ctx: Context) -> None:
...
+
+    @sudo.command()
+    @check(is_owner)
+    async def update(self, ctx: Context, package: str) -> None:
+        "Update a package"
+        async with ctx.typing():
+            os.system(f"pip install --upgraded {package}")
+        await ctx.message.add_reaction("✅")
+
+        await ctx.send(f"Updated {package}!")
```
ItsDrike, line 300: Similarly to review above, this doesn't make sense, we'll have access to the server this runs on, no need for a command like this.
```suggestion
```

ItsDrike (changes requested):

bot/cogs/common.py
```diff
@@ -178,6 +161,31 @@ async def thanks(self, ctx: Context, member: Member, *, reason: str = None) -> N
             embed.set_image(url="https://media.giphy.com/media/6tHy8UAbv3zgs/giphy.gif")
             await ctx.send(embed=embed)
 
+    @command()
+    @cooldown(1, 10, BucketType.user)
+    async def shorten(self, ctx: Context, *, link: str) -> None:
+        '''Makes a link shorter using the tinyurl api'''
```
ItsDrike, line 167: We use double quotes in our codebase, both in dosctrings and in regular strings when possible, fix this together with lines: 169, 173 and 177

ItsDrike (changes requested):

ItsDrike (approved):

PR #77, opened 2020-07-07
Add Verification

janaSunrise:


Pipfile
```diff
@@ -24,6 +24,8 @@ html2text = "*"
 googletrans = "*"
 html5lib = "*"
 py-stackexchange = "*"
+pre-commit = "*"
```
ItsDrike, line 27: Unacceptable, pre-commit is in `[dev-packages]` there is absolutely no reason to include it in `[packages]`

cogs/verification.py
```diff
@@ -0,0 +1,138 @@
...
+
+from discord import File, Error
+from discord.ext.commands import (
+    Cog,
+    Bot,
+    Context,
+    command,
+    CommandInvokeError,
+    has_permissions,
+)
```
ItsDrike, line 13: Consider using a better import structure:
```suggestion
from discord.ext.commands import (Bot, Cog, CommandInvokeError, Context,
                                  command, has_permissions)
```

cogs/verification.py
```diff
@@ -0,0 +1,138 @@
...
+                separators=(",", ": "),
+            )
+
+    @command()
+    async def verify(self, ctx: Context) -> None:
+        """Verify yourself (the bot will DM you)"""
+        try:
+            self.verify.use_count += 1
+        except AttributeError:
+            self.verify.use_count = 1
```
ItsDrike, line 61: This seems irrelevant, keeping track of verification should be done on a per-server basis, not on bot-basis and will be done with database (possibly by API)

cogs/verification.py
```diff
@@ -0,0 +1,138 @@
...
+
+        if self.verify.use_count % self.word_list_refresh_rate == 1:
```
ItsDrike, line 63: Simply use random to determine change. (Generate a simple random number, check the modulo == 0: is divisible) it's pretty non-standard to check modulo (remainder) is 1, instead, we should check 0 equality (divisibility)

ItsDrike (changes requested):

ActuallyDaneel (commented):
honestly Drike has said everything there is to say.

PR #62, opened 2020-06-26
Restructure file system

ItsDrike:
## Overview

We need to update our filestructure to a more scalable model.
Right now the file structure is getting very big and it's unpractical.

## TODO

* [x] Move every bot code related file into `bot/` directory
    - [x] Optimize imports to handle this
* [x] Rename `main.py` to `__main__.py` in order to be able to start it as a module
* [x] Merge all math-related files into single `cogs/math.py`
    - [x] Merge `octcalc.py`, `hexcalc.py` and `bincalc.py`
* [x] Remove `embed_handler.py` with only 1 function in it, referenced only once throughout the project
* [x] Remove `members.py` with only 1 function in it, referenced only once throughout the project
* [x] Remove unused `conversion.py`
* [x] Merge `cogs/utils` into `utils`
* [x] Merge `urbandictpages.py` into `study.py`
* [x] Move `constants.py` into `bot/`
* [x] Rename `constants.py` to `config.py`
* [x] Use `assets/filtered_words.txt` instead of defining these in python file
* [x] Rename `code_utils` to `coding` and use it as a module wiht `__init__.py` for cog loading and `coding.py` as the cog itself
* [x] Standardize hangman to use a separate class instead of defining everything in 1 command function
* [x] Split games cog into games/ directory which will contain games.py for smaller games and hangman.py for hangman
* [x] Use .txt file to store `allowed_filetypes` (AntiMalware) and `filter_words` instead of python files
* [x] Enable `malware_protection` cog
* [x] Rename `malware_protection.py` cog to `security.py`
* [x] Remove `search_utils/` and merge it's content directly into `search.py` cog
* [x] Remove `endpoints` and add a single dict that it contains into `constants.py`
* [x] Remove `emote_utils/` and merge its content directly into `emotes.py` cog
* [x] Split `coding.py` into `coding.py` and `documentation.py`
* [x] Optimize `coding.py`
* [x] Use proper imports for Bot child class from `bot/core/bot` instead of using `discord.ext.commands.Bot` for type hints
* [x] Move `assets/checks.py` to `utils/checks.py`
* [x] Create `core/converters.py` which will be used to store all converters in the project
* [x] Create `core/decorators.py` which will be used to store all decorators in the project
* [x] Cleanup `bot/utils/errors.py`
* [x] Rename `converters.py` cog to `conversion.py` to avoid confustion with `bot/core/converters.py`
* [x] Remove redundant `infog.py` cog

## Additional Notes

[...]

PR #78, opened 2020-07-08
Add some more Fun games

janaSunrise:


main.py
```diff
@@ -53,14 +53,14 @@ def __init__(self, *args, **kwargs) -> None:
...
-            )
-            self.change_status.start()
+            # self.pool = await asyncpg.create_pool(
+            #     database=os.getenv("DATABASE_NAME", "chaotic"),
+            #     host="127.0.0.1", min_size=int(os.getenv("POOL_MIN", "20")),
+            #     max_size=int(os.getenv("POOL_MAX", "100")),
+            #     user=os.getenv("DATABASE_USER"),
+            #     password=os.getenv("DATABASE_PASSWORD"),
+            # )
+            # self.change_status.start()
```
ItsDrike, line 63: Database keeping will be decided in #81, not here, we should leave it uncommented until that.
Same goes for the line 74

cogs/utils/constants.py
```diff
@@ -228,6 +228,713 @@
     default = safe_load(file)
 
 
+talk_games = {
```
ItsDrike, line 231: Unacceptable, use a `.txt` file in `assets` or direcrtly in `games/` (after #62) this is way too big for `constants.py`

Pipfile
```diff
@@ -16,7 +16,6 @@ aiohttp = "==3.6.2"
 beautifulsoup4 = "==4.9.1"
 discord = "==1.0.1"
 markdownify = "==0.4.1"
-stackexchange = "==0.5.0"
```
ItsDrike, line 19: Why is this a thing, `stackexchange` was fixed in #62

ItsDrike (changes requested):

PR #87, opened 2020-07-18
Added Dice and Coin flip

Rohith04MVK:


bot/cogs/games/games.py
```diff
@@ -39,3 +39,19 @@ async def ball8(self, ctx: Context, *, question: str) -> None:
         embed.add_field(name="Answer", value=answer)
         
         await ctx.send(embed=embed)
+    @command()
+    async def flip_coin():
+        def coin_filp():
```
ItsDrike, line 44: We require all functions to have proper docstrings as these are what will be displayed in the `>>help` even tho this command is pretty self-descriptive, it should still have a docstring, same goes fro `dice` command

bot/cogs/games/games.py
```diff
@@ -39,3 +39,19 @@ async def ball8(self, ctx: Context, *, question: str) -> None:
...
+            coin_filp = randint(0, 1)
+            if coin_filp == 0:
+                return "Heads!"
+            else:
+                return  "Tails"
+
+        await ctx.send(coin_filp())
+    from random import randint
```
ItsDrike, line 52: All import should be on top and properly sorted (accordingly to flake8)

ItsDrike (changes requested):

bot/cogs/games/games.py
```diff
@@ -39,3 +39,19 @@ async def ball8(self, ctx: Context, *, question: str) -> None:
         embed.add_field(name="Answer", value=answer)
         
         await ctx.send(embed=embed)
+    @command()
```
janaSunrise, line 42: you need to configure an alias for faster usage, instead of keeping just this long name
```suggestion
    @command(aliases=["cflip"])
```

bot/cogs/games/games.py
```diff
@@ -39,3 +39,19 @@ async def ball8(self, ctx: Context, *, question: str) -> None:
...
+            else:
+                return  "Tails"
+
+        await ctx.send(coin_filp())
+    from random import randint
+
+    @command()
+    async def dice():
+        dice_roll = randint(1,6)
+        await ctx.send(dice_roll)
```
janaSunrise, line 57: Give a default values for the range, and customizable range

janaSunrise (changes requested):

bot/cogs/games/games.py
e-xe: Actually, @ItsDrike we shouldn't worry about the `>>help` command since it will be replaced by the paginator once it is released.

ItsDrike: @e-exe even with it replaced docstrings will still be used as a command descriptions even with it being paginated

PR #67, opened 2020-07-01
Add Tic Tac Toe game

janaSunrise:

PR #88, opened 2020-07-19
Added leetify command and improved docstrings

ActuallyDaneel:
## Changes
- Implemented minor QOL changes on most of the docstrings, although one (the spoil docstring) didn't show up using the help command,
- Bugfix: `byteconvert` command sent 9 embeds due to indentation error
- Added `leetify` command: markdown spams a message, giving each letter a random text effect
- Bugfix: `hangman.py` had several syntax errors
- A lot of flake8 compliance commits (curse you azure pipeline)


janaSunrise (commented):
perfect, will be approved, once #67 is merged

bot/cogs/fun.py
```diff
@@ -199,6 +199,43 @@ async def lizard(self, ctx) -> None:
...
+    async def Leetify(self, ctx):
+        Leetters = {'a':['A','4'], 'b':['b', 'B'], 'c':['C', 'c'], 'd':['D', 'd'],
+                        'e':['E', '3'], 'f':['F', 'f'], 'g':['G', 'g'], 'h':['H', 'h'],
+                        'i':['1', 'I', 'i'], 'j':['j', 'J'], 'k':['K', 'k'], 'l':['L', 'l'],
+                        'm':['M', 'm'], 'n':['n', 'N'], 'o':['O', '0'], 'p':['P', 'p'],
+                        'q':['Q', 'q'], 'r':['R', 'r'], 's':['5', 'S'], 't':['t', 'T'],
+                        'u':['U', 'u'], 'v':['v','V'], 'w':['W', 'w'], 'x':['X', 'x'],
+                        'y':['y', 'Y'], 'z':['Z', 'z']}
+        content = [i.lower()+' ' for i in [i for i in ctx.message.content.split(" ") if i.lower() != ">leet" and i.lower() != ">leetify"]]
+        strContent = ''
```
ItsDrike, line 212: It is very non-standard to split variables with capital letter separation for a new word, instead, the pythonic way to do it would be:
```suggestion
        str_content = ''
```
This also applies to all variables like this.

bot/cogs/fun.py
```diff
@@ -199,6 +199,43 @@ async def lizard(self, ctx) -> None:
...
+                    MDlist.remove("*")
+                    MDlist.remove("**")
+                    MDlist.remove("***")
+                elif prevMD == '__':
+                    MDlist.remove("__")
+                elif prevMD == '`':
+                    MDlist.remove("`")
+                ChosenMD = random.choice(MDlist)
+                prevMD = ChosenMD
+                RefContent += ChosenMD+random.choice(Leetters[i])+ChosenMD
```
ItsDrike, line 233: We follow the usual `flake8` python syntax which requires whitespaces around arithmetic operators, so this line should look like this:
```suggestion
                RefContent += ChosenMD + random.choice(Leetters[i]) + ChosenMD
```

bot/cogs/fun.py
```diff
@@ -199,6 +199,43 @@ async def lizard(self, ctx) -> None:
...
+            else:
+                RefContent += i
+
+        await ctx.channel.send(RefContent + f"\n-{ctx.message.author.name}")
```
ItsDrike, line 237: doing `ctx.channel.send` will produce the same result as doing `ctx.send` except that's shorter so using that would be preferred (as it is everywhere throughout our codebase)
```suggestion
        await ctx.send(RefContent + f"\n-{ctx.message.author.name}")
```

ItsDrike (changes requested):
Nice work with the docstrings, they really needed some reformatting, but I did have a few requests for the proposed new command `leetify`.

We use double quotes (`"`) to signify a string everywhere throughout our codebase, please follow this standard in your contributions instead of using single quotes (`'`) and replace all your strings to comply with this.

bot/cogs/fun.py
therealdaneel: thanks, i'll not make this mistake moving forwards

ItsDrike:
@therealdaneel note that some of the suggestions I mentioned require further attention, simply committing them won't be enough in some cases as they might also apply to other variables, etc which I haven't pointed out since that would be too long, please read their individual description and resolve them in an editor, after that you can simply push it to the same branch.

Also, note that you're currently working on `master` branch of your [forked repository](https://github.com/therealdaneel/HotWired-Bot) you shouldn't be doing that as you might need to get your branch up to date with ours and working on `master` will only cause many `pull` and `merge` commits which could be avoided by working on a custom branch (as described in our [`CONTRIBUTING.md`](https://github.com/The-Codin-Hole/HotWired-Bot/blob/master/CONTRIBUTING.md)) in the 2. section

image

I've noticed you've also made multiple mistakes regarding our contribution guidelines which I've pointed out, but to avoid further confusion with our code-style and other guidelines that we respect, make sure to properly read the CONTRIBUTING.md file, I know it's a bit long but it really describes how to contribute in great detail and avoiding these mistakes is the purpose of that file.

bot/cogs/fun.py
```diff
@@ -199,6 +199,43 @@ async def lizard(self, ctx) -> None:
                 else:
                     await ctx.send(f"Something went Boom! [status : {lizr.status}]")
 
+    @commands.command(aliases=["leet"])
+    async def Leetify(self, ctx):
```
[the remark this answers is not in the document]
ItsDrike: Purely commiting this won't help you as you're overwriting the `context` below this, you'll need to actually adjust the code you're using to use this `content` instead of getting it from your list comprehention:
```py
[i for i in ctx.message.content.split(" ") if i.lower() != ">leet" and i.lower() != ">leetify"]
```

bot/cogs/fun.py
ItsDrike: As I've already said:
> **This also applies to all variables like this and they need to be renamed accordingly.**

So please, don't mark it as resolved until those are fixed too, also note that by directly committing this you've only changed the definition of `strContent` to `str_content` but not the references to that variables, which will result in an error.

bot/cogs/fun.py
```diff
@@ -199,6 +199,43 @@ async def lizard(self, ctx) -> None:
...
+        Leetters = {'a':['A','4'], 'b':['b', 'B'], 'c':['C', 'c'], 'd':['D', 'd'],
+                        'e':['E', '3'], 'f':['F', 'f'], 'g':['G', 'g'], 'h':['H', 'h'],
+                        'i':['1', 'I', 'i'], 'j':['j', 'J'], 'k':['K', 'k'], 'l':['L', 'l'],
+                        'm':['M', 'm'], 'n':['n', 'N'], 'o':['O', '0'], 'p':['P', 'p'],
+                        'q':['Q', 'q'], 'r':['R', 'r'], 's':['5', 'S'], 't':['t', 'T'],
+                        'u':['U', 'u'], 'v':['v','V'], 'w':['W', 'w'], 'x':['X', 'x'],
+                        'y':['y', 'Y'], 'z':['Z', 'z']}
```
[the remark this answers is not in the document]
ItsDrike: As I've pointed out in the first section:
> **The same goes for every other capitalized variable name in this function, it need to be lowercased too.**

Don't mark this as resolved until that is resolved too.

Also note that below the suggestion I've proposed a way of avoiding this huge dictionary and only keep track of needed keys, not every letter in alphabet which is often the same. Please read through the suggestion properly and only after you've completed everything requested mark it as resolved.

bot/cogs/fun.py
ItsDrike: This suggestion also applies to all occurrences of breaking this standard flake8 syntax concerning whitespaces around these operator, for example you also violate it on line 213 with the complicated list comprehention:

```py
 content = [i.lower()+' ' for i in [i for i in ctx.message.content.split(" ") if i.lower() != ">leet" and i.lower() != ">leetify"]]
```

These conventions need to be followed everywhere, the suggestion was only 1 example of it

bot/cogs/fun.py
```diff
@@ -199,6 +199,47 @@ async def lizard(self, ctx) -> None:
...
+        leetters = {
+            "a": ["A","4"], "b": ["b", "B"], "c": ["C", "c"], "d": ["D", "d"],
+            "e": ["E", "3"], "f": ["F", "f"], "g": ["G", "g"], "h": ["H", "h"],
+            "i": ["1", "I", "i"], "j": ["j", "J"], "k": ["K", "k"], "l": ["L", "l"],
+            "m": ["M", "m"], "n": ["n", "N"], "o": ["O", "0"], "p": ["P", "p"],
+            "q": ["Q", "q"], "r": ["R", "r"], "s": ["5", "S"], "t": ["t", "T"],
+            "u": ["U", "u"], "v": ["v","V"], "w": ["W", "w"], "x": ["X", "x"],
+            "y": ["y", "Y"], "z": ["Z", "z"]
+        }
+        content = [i.lower() + " " for i in ctx.message.content[1:]]
```
ItsDrike, line 215: Since you're getting `content` as an argument of that command now, there is no reason to use `ctx.message.content[1:]` it's exactly the same as doing `content` except `content` is more convenient.

```suggestion
        content = [i.lower() + " " for i in content]
```
also note that converting content to lowercase can be done in a much easier way like this:
```suggestion
        content = content.lower()
```
this can be done since `lower()` works for multi-letter strings too and does what you did automatically.

bot/cogs/fun.py
```diff
@@ -199,6 +199,47 @@ async def lizard(self, ctx) -> None:
...
+
+        leetters = {
+            "a": ["A","4"], "b": ["b", "B"], "c": ["C", "c"], "d": ["D", "d"],
+            "e": ["E", "3"], "f": ["F", "f"], "g": ["G", "g"], "h": ["H", "h"],
+            "i": ["1", "I", "i"], "j": ["j", "J"], "k": ["K", "k"], "l": ["L", "l"],
+            "m": ["M", "m"], "n": ["n", "N"], "o": ["O", "0"], "p": ["P", "p"],
+            "q": ["Q", "q"], "r": ["R", "r"], "s": ["5", "S"], "t": ["t", "T"],
+            "u": ["U", "u"], "v": ["v","V"], "w": ["W", "w"], "x": ["X", "x"],
+            "y": ["y", "Y"], "z": ["Z", "z"]
+        }
```
ItsDrike, line 214: As I've pointed out in a review before, I'm personally not a fan of this dictionary since in most of the cases it simply follows this: `letter: [letter, letter.upper()]`, we don't need to repeat ourselves here and generate a dictionary for every single letter in the alphabet, we can make it easier and only define the special cases like `"a": ["A", "4"]` and for those which won't be there simply default it to a random choice of that same letter or uppercase version of it.
It would look something like this:
```py
leetified_word = ""
# In python you can loop through individual letters by doing `for letter in string`
for letter in content:
    # The letter is in the dictionary (that means it is one of the special cases where we don't only use random choice of lowercase and uppercase version of that letter)
    if letter in letters:
       leetified_word += random.choice(leeters[letter])
    # If the letter isn't defined in the leeters, use the upper or lowercase version randomly
    else:
        leetified_word += random.choice([letter, letter.upper()])
```
This suggestion is basically a much simpler version of most of the proposed function

bot/cogs/github.py
```diff
@@ -20,10 +20,10 @@ def __init__(self, bot: Bot) -> None:
 
     @command(aliases=["pullrequest", "pullrequests", "issues"])
     @cooldown(1, 5, type=BucketType.user)
-    async def issue(self, ctx: Context, number: int, repository: str = "HotWired-Bot", user: str = "The-Codin-Hole") -> None:
+    async def issue(self, ctx: Context, issueNum: int, repository: str = "HotWired-Bot", user: str = "The-Codin-Hole") -> None:
```
ItsDrike, line 23: As described before, it's uncommon to use capital letters as word separation in python, please change this to `issue_number` and also update all references to it

ItsDrike (changes requested):

bot/cogs/fun.py
ItsDrike: You've committed the first suggested part, but as described below that, you don't actually need a list comprehension and convert each individual letter of the string into a list, you can simply use `.lower()` on the whole, string which will make all letters in it lowercase and use `list()` around it which will convert the string into a list of its words.

This is the preferred approach as `lower` is a built-in method running in `C` language instead of in python, which provides a faster way to convert string to lowercase instead of re-implementing it by that list comprehension

bot/cogs/fun.py
therealdaneel: thanks to your review, i've shortened the dictionary length to 5 items.

bot/cogs/fun.py
```diff
@@ -199,6 +199,43 @@ async def lizard(self, ctx) -> None:
...
+
+        leetters = {
+            "a": ["A","4"], "e": ["E", "3"], "i": ["1", "I", "i"],
+            "o": ["O", "0", "o"], "s": ["5", "S", "s"]
+            }
+        content = [i.lower() + " " for i in content]
+        str_content = ""
+        for i in content:
+            str_content += i
+        ref_content = ""
```
ItsDrike, line 214: This seems like a better name for this.
Do note that you will need to update all references to `ref_content` with this new name.
```suggestion
        leetified_content = ""
```

ItsDrike (changes requested):
If every requested change is addressed properly, this should be ready to be merged.

ItsDrike:
> perfect, will be approved, once #67 is merged

@janaSunrise I don't see a reason to wait for #67 this PR doesn't seem isn't dependant on it so it should be mergeable before that. Once the proposed changes by me will be addressed it should be ready for merge (that is if you'll not have any further suggestions).

bot/cogs/fun.py
```diff
@@ -199,6 +199,43 @@ async def lizard(self, ctx) -> None:
...
+                md_list.remove("*")
+                md_list.remove("**")
+                md_list.remove("***")
+            elif prev_md == "__":
+                md_list.remove("__")
+            elif prev_md == "`":
+                md_list.remove("`")
+            chosen_md = random.choice(md_list)
+            prev_md = chosen_md
+            ref_content += chosen_md + leet_char + chosen_md
```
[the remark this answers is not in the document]
janaSunrise: basically, faster concatenation

bot/cogs/search.py
```diff
@@ -130,7 +130,15 @@ async def _basic_search(self, ctx, query: str, category: str) -> None:
...
-        """Search online for general results."""
+        """Search online for general results.
+           Valid Categories:
+                            - web
+                            - videos
+                            - music
+                            - files
+                            - images
+                            - it 
+                            - maps"""
```
ItsDrike, line 141: Minor formatting adjustments
```suggestion
        """
        Search online for general results.
        Valid Categories:
             - web
             - videos
             - music
             - files
             - images
             - it 
             - maps
        """
```

bot/cogs/fun.py
```diff
@@ -199,14 +199,54 @@ async def lizard(self, ctx) -> None:
...
     @command()
     async def why(self, ctx: Context) -> None:
         """Why?."""
         async with aiohttp.ClientSession() as session:
             async with session.get("https://nekos.life/api/why", headers=self.user_agent) as why:
                 if why.status == 200:
                     why_js = await why.json()
-                    why_em = discord.Embed(title=f"{ctx.author.name} wonders...", description=why_js["why"], colour=0x690E8,)
+                    why_em = discord.Embed(
+                        title=f"{ctx.author.name} wonders...", description=why_js["why"], colour=0x690E8,)
```
ItsDrike, line 249: We should use this formatting when splitting an embed, it is more readable.
We also don't need to use `discord.Embed` as we have `Embed` imported directly
```suggestion
                    why_em = Embed(
                        title=f"{ctx.author.name} wonders...", 
                        description=why_js["why"], 
                        colour=0x690E8
                    )
```

ItsDrike (changes requested):
Minor formatting adjustments

bot/cogs/fun.py
ItsDrike: Hey, you didn't actually resolve this one, please avoid marking unresolved requests as resolved.

therealdaneel: I've gone through the embeds, is there any particular reason for a trailing comma at the embed constructors arguments end?
`Embed(title="title", color=color.gold(),)`

bot/cogs/commands.py
```diff
@@ -52,8 +53,7 @@ async def serverinfo(self, ctx: Context) -> None:
     async def userinfo(self, ctx: Context, user: t.Optional[t.Union[Member, User]] = None) -> None:
         """
         Get information about you, or a specified member.
-
-        `user`: The user to get information from. Can be a Mention, Name or ID.
+        'user' can be a user Mention, Name, or ID.
```
ItsDrike, line 56: Actually we insist on using \`user\` instead of 'user' as \` is a symbol which marks inline code in discord markdown and this help docstring will be displayed in a help command for this function so we should keep the user wrapped in backticks instead of single quotes

ItsDrike (changes requested):
Merge our version of `master` into this so that you resolve the conflict with `sudo.py` and it will also remove the `flake8-import-order`

therealdaneel:
sudo.py is throwing errors, i have no idea why, as I didn't touch it

ItsDrike:
@therealdaneel The sudo has a merge conflict in it, that's the reason why it's throwing errors, your version of sudo wasn't compatible with our version and it produced a merge conflict, you commited the file with that merge conflict:
https://github.com/The-Codin-Hole/HotWired-Bot/blob/f622de5fdf5d0bc9ce076a4964918596706b4b85/bot/cogs/sudo.py#L10-L18

ItsDrike (approved):
Good to go

janaSunrise (approved):

PR #89, opened 2020-07-21
Added pokemon command

ActuallyDaneel:
searches up a given pokemon's name using `pokeAPI`.
Displays the following info: 
- base stats
- type
- weight
- abilities.

### DIFF
--- a/bot/cogs/comics.py
+++ b/bot/cogs/comics.py
@@ -13,7 +13,7 @@ from discord.ext.commands import Cog, Context, command
 
 
 class Comics(Cog):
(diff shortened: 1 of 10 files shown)


ItsDrike:
For some reason, in this PR you've implemented everything from #88 and added something on top of it, what you should normally do when making new PRs is to make a new branch which diverges from the last commit in our `master` branch and follow from there, not making a branch with the changes of #88 already included, that way you're only re-implementing what's already there, for now, this will need to wait until #88 is merged so that this won't re-implement its changes

PR #91, opened 2020-07-21
Add Prefixes that can be changed on a server basis

janaSunrise:


ghost (changes requested):
This is not a proper way of doing this, we should wait for a database

ActuallyDaneel (changes requested):
A prefix length limit should be implemented to prevent someone from cluttering the json with a 2000 char prefix (i'm thinking 15-20 characters max)

ItsDrike:
> This is not a proper way of doing this, we should wait for a database

@e-xe 
We will need to implement this as another API request which will save it into its internal database, that way we can extract or change this prefix easily from the web pannel as it can also make API requests.
Implementing a full database for the bot isn't necessary due to the presence of API

ItsDrike:
> A prefix length limit should be implemented to prevent someone from cluttering the json with a 2000 char prefix (i'm thinking 15-20 characters max)

Even though we won't be using `json` to store these in the future I still agree, we should define a maximum length for the prefix.

PR #92, opened 2020-07-21
Remove flake8-import-order and install flake8-bugbear

ItsDrike:
- Remove `flake8-import-order` from Pipfile and pre-commit-config.yaml as it isn't compatible with our style of import order, we can design a proper way to do this later
- Install `flake8-bugbear` to catch some bad code practices

### DIFF
--- a/.azure-pipelines.yaml
+++ b/.azure-pipelines.yaml
@@ -16,11 +16,11 @@ jobs:
         displayName: "Installing dependencies"
 
       - script: |
-          python -m pipenv run lint
-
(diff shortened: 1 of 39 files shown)


janaSunrise (approved):

janaSunrise (dismissed):

ItsDrike:
This is a critical PR since `flake8-import-order` currently causes problems with all other PRs, until that's fixed I've removed it, 
I'm overriding common practices to wait for approvals as this needs to be done now, otherwise, all other PRs are pending

PR #97, opened 2020-07-24
added documentation command

Anubhav1603:
this command has Rapptz fuzzy finder and able to search documentation like R.danny

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -28,7 +28,7 @@ lxml = "~=4.5.2"
 bleach = "~=3.1.5"
 
 [requires]
-python_version = "3.8"
+python_version = "3"
 
 [scripts]
 start="python -m bot"
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -1,9 +1,11 @@
 import os
+import asyncio
 
 from discord import Game
 from discord.ext.commands import when_mentioned_or
 
 from bot import config
+from aiohttp import ClientSession
 from bot.core.bot import Bot
 
 TOKEN = os.getenv("BOT_TOKEN")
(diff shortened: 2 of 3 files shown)


Anubhav1603:
i tried to make changes in code but if i remove if part or else part its error out

therealdaneel:
You need to check your changes against flake8 and resolve those conflicts because this 
image
is why your code is failing azure pipelines

ActuallyDaneel (changes requested):
Looks like you copy/pasted a command from somewhere, in order to make it merge-ready, you'll have to make a few changes

- add `documentation.py` to the extension list in `__main__.py`
- resolve flake8 errors/warnings
- add docstrings to all the functions and classes
- add typehints to all function arguments

ItsDrike (changes requested):
As was already pointed out, resolve all `flake8` issues found by Azure Pipelines and read [`CONTRIBUTING.py`](https://github.com/The-Codin-Hole/HotWired-Bot/blob/master/CONTRIBUTING.md) which contains the formatting rules for our codebase, please make sure to follow them in your code.

Also, note that we don't accept copied code from other bots, so things like `"Credits to github.com/Rapptz"` are unacceptable, you'll need to write your own non-copied functions as we require original non-copied codebase (If there is something copied, it was done with authors permission not to mention them and it's usually just some smaller parts of code), please respect that.

Anubhav1603:
can i share my code which i use ItsDrike for other commands

therealdaneel:
> can i share my code which i use ItsDrike for other commands

if it's your own original code, then you can use it. ItsDrike's issue was that you were copy/pasting code you didn't have permission to use.

PR #98, opened 2020-07-24
Make config an extension

Faholan:
That way we'd be able to use botvars everywhere instead of each time having to import

### DIFF
--- a/bot/__main__.py
+++ b/bot/__main__.py
@@ -8,34 +8,7 @@ from bot.core.bot import Bot
 
 TOKEN = os.getenv("BOT_TOKEN")
 
-extensions = [
-    "bot.cogs.codesandbox",
-    "bot.cogs.commands",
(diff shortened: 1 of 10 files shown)


ActuallyDaneel (commented):
why did you change every single `Embed()` call to `discord.Embed()`?

Faholan:
> 
> 
> why did you change every single `Embed()` call to `discord.Embed()`?

Because it was imported twice, once all alone and once in discord. Plus there were some instances of discord.Embed mixed with the Embed.

ItsDrike (commented):
To be honest, I'm not a huge fan of this, why do we need to have config parameters added to `bot` directly, this adds way too many parameters to `bot` which in my opinion doesn't make sense as they doesn't have anything to do with `bot`. I think having them in `config` alone makes more sense, but this is only my opinion and I alone can be overvoted.

PR #114, opened 2020-08-05
Update README.md

Sadashii:


LostSoul0:
The old README had a discord link to the old server. So updated!

